PR #7, opened 2016-02-10
Fix package name in installation instructions

zeppelin:


machty:
Missed this, already made the change :)

PR #12, opened 2016-02-18
Enable YUIDoc generation.

rwjblue:
This adds the following commands:

```
ember ember-cli-yuidoc
ember serve --docs
```

Both commands will emit YUIDoc generated JSON and HTML output in `./docs` folder, which can then be published to the website if desired.

There are a number of themes that we can investigate to make the default generated HTML nice enough to include on the site, but in the meantime fleshing out the actual API docs is a great way to make the code base more accessible to newbs (like myself)...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ TODO.md
 
 deploy.sh
 
+/docs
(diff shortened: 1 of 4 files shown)


rwjblue:
Current HTML output looks like:

default look/feel

PR #36, opened 2016-03-17
Added wait hover example

rtablada:
This example shows how ember-concurrency could be used for waiting for hover events.

This could be a common use case in hover UI's since the naive approach for hover is to immediately activate.
But, when thinking about waiting for hovers to be confirmed and the user to stay on an element,
this becomes REALLY hard to manage the timers and canceling.

Luckily, with `ember-concurrency` this is REALLY easy to support.

This could be used for things like showing dynamic tool tips, or dropdowns that wait before activating.

### DIFF
--- a/tests/dummy/app/components/hover-button/component.js
+++ b/tests/dummy/app/components/hover-button/component.js
@@ -1,0 +1,18 @@
+import Ember from 'ember';
+
+// BEGIN-SNIPPET hover-button
+function sendHover() {
+  this.sendAction('hover');
+}
+
+function sendRelease() {
+  this.sendAction('release');
+}
+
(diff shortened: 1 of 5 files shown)


machty:
Thanks for the PR :)

I think for this to be mergeable it would need to actually include a real dropdown menu. It's not obvious what's happening just from hovering the button, and it'd show it's power more if it were an actual menu (kind of like we demonstrate a real debounced type-ahead search). What do you think?

rtablada:
Yeah I can do that. Wanted to get the primitive done first to see if I could do it. Give me a bit and I'll see what I can do about the dropdown.

machty:
Closing this til there's an update.

PR #65, opened 2016-07-24
remove ember-try from package to fix deprecation

BryanCrotaz:
ember-try is now included in ember-cli so remove from package.json


machty:
The CI build's breaking; you'll probably also need to bump the ember-cli version for tests to pass.

PR #70, opened 2016-08-17
Exclude rx.js from npm package

jakesjews:


machty:
Can you remind me why this is important to remove? Is it causing problems / bloat from consuming applications?

jakesjews:
@machty just trying to reduce the amount of stuff downloaded with an NPM installed based on the overall story https://github.com/ember-cli/ember-cli/issues/6147. If you would rather keep it in the package I have no problem closing the PR.

jakesjews:
If you use git for pulling down the test code you should be ok. This just excludes it from the published npm package.

machty:
I did some cleanup and ended up removing this and some other things from the npm package. Thanks for the heads up :)

PR #73, opened 2016-09-14
Clarify timeout method usage

alvincrespo:
- Helps clarify where the timeout method is being imported from for first time users implementing their first task

At first I thought adding a comment with a prefix of `//NOTE: ` would help, but thinking it through - I believe exposing the whole script would make it clearer for those looking to just try out the task without having to reference the API.


alvincrespo:
Hmm. Not sure why I'm getting these kind of failures:

```
not ok 180 Chrome 53.0 - TestLoader Failures: dummy/tests/acceptance/root-test: could not be loaded
```

Issue #74, opened 2016-09-21
allSettled task aware yieldable


aaronbhansen:
Right now you have an "all" yieldable helper you're exporting that takes advantage of the taskAwareVariantOf method. Would it make sense to add a allSettled yieldable so we can handle the errors after all promises have completed? If so I could add a quick pull request to add it.

machty:
Yeah, that sounds like a winner to me. Go for it :)

machty:
Can you reproduce this travis failure locally?

<details>
  <summary>Failing travis errors</summary>
not ok 194 PhantomJS 2.1 - Unit: yieldables: allSettled does not cancel all other joined tasks if one of them fails
    ---
        actual: >
            Max depth.,Max depth.,Max depth.
        expected: >
            Max depth.,Max depth.,Max depth.
        stack: >
            callee$1$0$@http://localhost:7357/assets/tests.js:4211:31
            tryCatch@http://localhost:7357/assets/vendor.js:3823:44
            invoke@http://localhost:7357/assets/vendor.js:4090:30
            http://localhost:7357/assets/vendor.js:3856:28
            _takeSafeStep@http://localhost:7357/assets/vendor.js:103014:45
            _takeStep@http://localhost:7357/assets/vendor.js:103029:46
            invoke@http://localhost:7357/assets/vendor.js:14255:21
            flush@http://localhost:7357/assets/vendor.js:14321:17
            flush@http://localhost:7357/assets/vendor.js:14122:22
            end@http://localhost:7357/assets/vendor.js:14500:30
            run@http://localhost:7357/assets/vendor.js:14622:21
            run@http://localhost:7357/assets/vendor.js:34600:32
            http://localhost:7357/assets/tests.js:4259:26
            runTest@http://localhost:7357/assets/test-support.js:2688:32
            run@http://localhost:7357/assets/test-support.js:2673:11
            http://localhost:7357/assets/test-support.js:2815:14
            process@http://localhost:7357/assets/test-support.js:2474:24
            begin@http://localhost:7357/assets/test-support.js:2456:9
            http://localhost:7357/assets/test-support.js:2516:9
        Log: |
    ...
</details>

aaronbhansen:
I tried in phantomjs and just got the error. I was running all the tests in chrome. I'll see if I can fix it in phantomjs.

aaronbhansen:
[...]

PR #75, opened 2016-09-21
Adding allSettled to yieldable operations

aaronbhansen:
This adds allSettled as an available yieldable operation for the issue #74

### DIFF
--- a/tests/unit/yieldables-test.js
+++ b/tests/unit/yieldables-test.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import { task, all } from 'ember-concurrency';
+import { task, all, allSettled } from 'ember-concurrency';
 
 module('Unit: yieldables');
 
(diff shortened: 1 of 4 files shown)


tests/unit/yieldables-test.js
```diff
@@ -111,3 +111,114 @@ test("all cancels all joined tasks if parent task is canceled", function(assert)
...
+    parent: task(function * () {
+      let task = this.get('child');
+      let allPromise = allSettled([
+        task.perform(),
+        task.perform(),
+        task.perform(),
+      ]);
+      assert.equal(typeof allPromise.then, 'function');
+      let values = yield allPromise;
+      assert.deepEqual(values, [{ state: 'fulfilled', value: 'a' }, { state: 'fulfilled', value: 'b' }, { state: 'fulfilled', value: 'c' }]);
```
machty: if Phantom is failing this, the least hacky solution might be `Ember.A([values]).map(({state, value}) => ({ state, value }))`

aaronbhansen: Here is the full log of the values, the stack and line are under the reason, thats why I built out the other solution to re-construct reason object. https://cl.ly/1c1E2u0w1H3O

Issue #77, opened 2016-09-27
`.maxConcurrency()` on its own should throw an error


machty:
If you just do `task(...).maxConcurrency(n)` without applying any other modifier, we should probably raise an error and tell you to use one of the other task modifiers for clarity. This would be a "breaking" change but it would nudge you to clarify intent, whereas by default today `.maxConcurrency()` on its own behaves the same as `.enqueue().maxConcurrency(n)`, which has caused some confusion because some people think it implies `.drop()`, others `.restartable()`, etc.

ddoria921:
Gonna take a shot at this

PR #79, opened 2016-09-30
[Issue #77] maxConcurrency called without specifying a task modifier

ddoria921:
## What is this?
If you do ` task(...).maxConcurrency(n)` without applying any other modifier it's not really obvious what the intent is and has led to some confusion. In an effort to clear things up this PR will make it so that a warning (possibly signaling a deprecation) is logged in this scenario.

## What I Did
I created two new flags in the `-property-modifiers-mixin`. One is to track if a modifier was set and the other to track if `maxConcurrency()` was called. Then, in the `setup` hook I check to see if `maxConcurrency()` was called without setting a modifier. If it was I log a warning.

### DIFF
--- a/addon/-task-property.js
+++ b/addon/-task-property.js
@@ -408,6 +408,10 @@ objectAssign(TaskProperty.prototype, propertyModifiers, {
   constructor: TaskProperty,
 
   setup(proto, taskName) {
+    if (this._hasSetMaxConcurrency && !this._hasSetBufferPolicy) {
+      Ember.Logger.warn(`The use of maxConcurrency() without a specified task modifier is deprecated and won't be supported in future versions of ember-concurrency. Please specify a task modifier instead, e.g. \`${taskName}: task(...).enqueue().maxConcurrency(${this._maxConcurrency})\``);
+    }
+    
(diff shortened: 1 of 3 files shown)


addon/-task-property.js
```diff
@@ -408,6 +408,10 @@ objectAssign(TaskProperty.prototype, propertyModifiers, {
   constructor: TaskProperty,
 
   setup(proto, taskName) {
+    if (this._hasSetMaxConcurrency && !this._hasSetBufferPolicy) {
```
machty: Maybe instead of tracking `_hasSetMaxConcurrency` you just replace it with a check of `this._maxConcurrency === Infinity`? The fewer the properties to keep track of the better, and now that I think about it it's probably the most logical thing we're trying to prevent (the idea that when MC is infinity, all the modifiers behave the same). For reference, [this is where the default maxConcurrency is set to Infinity](https://github.com/machty/ember-concurrency/blob/72f70b6c327f5242ca623d61ea0595b5f9093896/addon/-property-modifiers-mixin.js#L13).

ddoria921: I thought about checking for Infinity but figured the warning wouldn't get logged if you manually set it to Infinity. The more I think about it... people probably aren't doing that. 

I'll make the change.

PR #82, opened 2016-10-14
Configure task property definitions (read/write)

### LINKED ISSUE #81: Task overriding broken in 0.7.11
In 0.7.10, overriding a task definition in an `extend` call worked as I'd expect a normal property override to, but with the getter changes in 0.7.11 it now throws `TypeError: Cannot set property go of [object Object] which has only a getter`.

```js
const MyThing = Ember.Object.extend({
  go: task(function*() { return 'yes'; })
});

const MyOtherThing = MyThing.extend({
  go: task(function*() { return 'actually, no'; })
});
```

sukima:
Make `configurable` and `writable`.

Closes #81

### DIFF
--- a/addon/-task-property.js
+++ b/addon/-task-property.js
@@ -418,6 +418,8 @@ objectAssign(TaskProperty.prototype, propertyModifiers, {
 
     let cp = this;
     Object.defineProperty(proto, taskName, {
+      configurable: true,
+      writable: true,
       get() {
         return cp.get(this, taskName);
       }


machty:
@sukima thank you for the PR; Could you please provide a breaking test though? It should be reasonably easy to write below [this test](https://github.com/machty/ember-concurrency/blob/master/tests/unit/task-test.js#L237) and following a similar style. And this goes with out saying but make sure it actually breaks w the error reported in #81.

machty:
Also does it need to be writable? or just configurable?

dfreeman:
Aren't property descriptors either `writable`/`value` or `get`/`set`? What happens when you mix'n'match?

machty:
subsumed by #83

PR #90, opened 2016-10-31
Use `Ember.ComputedProperty` instead of private modules.

rwjblue:


machty:
Huh, so this has been exposed on `Ember` for at least 1.13?

rwjblue:
Yep: https://github.com/emberjs/ember.js/blob/release-1-13/packages/ember-metal/lib/main.js#L328

PR #93, opened 2016-11-09
Route tasks example: add this._super to setupController hook

mike-north:


machty:
I feel like I should know this, but... is it necessary / good practice to call super() in setupController?

mike-north:
If it's not, [this code](https://github.com/emberjs/ember.js/blob/306b88fd60739ae8179d8fdb9b3a2b53c5112a71/packages/ember-routing/lib/system/route.js#L1765-L1769) will not be invoked AFAIK

machty:
I don't have strong opinions about this but seem fine.

machty:
Also, thanks1

Issue #54, opened 2016-05-16
Scheduler doesn't match async-await timing


machty:
At present, all `perform()`s are deferred to the `actions` queue, which occasionally produces different results from putting the same code in an `Ember.on()` function, e.g.

``` js
export default Ember.Component.extend({
  afterInit: Ember.on('init', function() {
    // runs synchronously during/after `init`
    this.set('wat', 'lol');
  }),
  // vs
  afterInitTask: task(function * () {
    // happens later, on the `actions` queue
    this.set('wat', 'lol');
  }).on('init'),
});
```

I noticed this was causing undesired animations to occur with Liquid Fire which were fixed by changing the code back to user `Ember.on('init')`. I don't know what the solution is now, but given that most libraries assume that logic in hooks is performed synchronously when the hook is called, ember-concurrency should probably make it possible to perform logic synchronously. Or perhaps e-c can maintain a microqueue that it trampoline-flushes every time `.perform()` is called.

machty:
Here's the plan: ember-concurrency's gonna get a new scheduler to bring it closer in line with how async-await timing works.

```js
    async function wat() {
      console.log(1);
      await null;
      console.log(3);
    }

    wat();
    console.log(2);
```

This produces 1 2 3

```js
    myTask: task(function * () {
      console.log(1);
      yield null;
      console.log(3);
    }),

    myTask.perform();
    console.log(2);
```

Today, this produces 2 1 3 because `.perform()` doesn't immediately (synchronously) start executing your task until the actions run loop queue. This is 1) inconsistent w async-await semantics, which is bad because ember-concurrency strives to match async-await semantics, but with the addition of cancelation, and 2) problematic with `.on('init')` or any time you perform a task from some third party API hook that expects some side effect to be performed before control is returned to the API.

So I'm going to change the scheduler to make the second example above print 1 2 3.

machty:
## To Sync, or not To Sync

After talking over w the core team, I don't think we can/should make all tasks begin execution synchronously by default. Instead, we should make it possible to opt into synchronous execution.

The reason we can't make synchronous start the default is because we can't guarantee synchronous execution in the case of restartable tasks.

```js
   foo: task(function * () {
     yield timeout(100);
     this.get('foo').perform(); // what happens here?
   }).restartable()
```

[...]

PR #100, opened 2016-12-04
Re-write scheduler to be more eager/synchronous

machty:
## EC Scheduler Sync

Since its conception, ember-concurrency's internal scheduler (the thing that executes the task generator functions, interprets yielded promise-y values and determines when/how a task function should resume with a yielded value) has been all-asynchronous. Specifically, anything that impacted the execution of a task (perform(), cancel(), or a yielded promise) would be deferred to a run loop queue (just like how RSVP promises defer callbacks to run on the `actions` queue, which makes perfect sense). For example, `.perform()`ing a task wouldn't actually start executing the task until the 'actions' run loop queue ran, [which is at odds with how async functions execute](https://github.com/machty/ember-concurrency/issues/54), which means timing-dependent hooks (like a lot of DOM APIs, or Ember lifecycle hooks (e.g. `.on('init')`) couldn't make use of Tasks without missing a crucial synchronous window.

This PR attempts to bring EC closer in line with async function scheduling, but also in general to make task execution more synchronous... but only when it's safe to do so. In particular:

1. `.perform()` starts synchronously executing the task if it can (e.g. it doesn't get `drop()`ped).
2. RSVP promises that synchronously resolve will synchronously(-ish) resume a task that, for example, does a `yield RSVP.Promise.resolve(123)`. This takes advantage of the fact that RSVP promises synchronously expose their resolution state/values, which EC can use to its advantage to do things more eagerly and with less overhead (see below if this sounds like "zalgo").

## Beginnings of Yieldable API

The beginnings of a "Yieldable" API: we expose a Symbol that 3rd party libraries can use to implement special behavior that should occur when an object is yielded in an EC task. This makes it possibly for third parties to implement special logic for when values are yielded in an EC task. In addition to simplifying EC internals for how values are yielded, task instances wired up, cancelation events propagated, etc., this is meant to open the door to community experimentation of async concepts/ideas without forcing folk to learn EC's complicated internals.

The eager/synchronous RSVP Promise resume behavior described above is implemented using this API.

## Zalgo?

http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony

TL;DR "Zalgo" describes the chaos that occurs when apps/libraries/frameworks provide APIs that are sometimes synchronous, sometimes asynchronous. Chiefly, this refers to APIs that are callback-based. As the various articles on Zalgo describe, it's _very_ difficult to use an API that sometimes invokes your callbacks synchronously, sometimes invokes them asynchronously. The best practice is if you can't guarantee one or the other, make the callbacks ALWAYS asynchronous.

Promise libraries consistent with the A+ spec (including RSVP) all work this way (as does native window.Promise). So is what I'm describing above a red flag?

_Possibly_, but probably not. Why not? Because the linear / flattened style of async functions and EC generator task functions "absorbs" the Zalgo. By `await`ing / `yield`ing a possibly async value, you unify how the function proceeds with the resolved value, whether it resolves synchronously or asynchronously in a manner that callback-based APIs fundamentally can't.

There are other use cases where APIs that can absorb zalgo can more eagerly do the right thing with synchronous values; [Handlebars templates are another example](https://twitter.com/wycats/status/804518510093139969).


machty:
I'm gonna need some help testing out these changes in advance since there is some risk that these subtle changes in scheduling might impact/break existing apps (but probably/hopefully only those apps that were doing questionable things).

machty:
Forgot to mention: all of the callback-based APIs, like `.then()` will continue to fire asynchronously. I'm just changing the behavior for when you yield possibly-synchronously-completing tasks/promises.

machty:
Anyone looking to try out the changes in this PR can use the the `pr-100` npm dist tag, or use the following twiddle:

https://ember-twiddle.com/febf2fe17782f6d08e0df16f951aa6d0?openFiles=controllers.application.js%2C

machty:
Also, this PR introduces a lot of improvements to performance (partially spurred on by continued experiments to integrate liquid fire with EC)

machty:
Example of how you can use the Yieldables API to build CSP coroutines https://twitter.com/machty/status/805591068301553665

machty:
Closing in favor of #107

PR #103, opened 2016-12-07
Update to ember-getowner-polyfill@1.1.0.

rwjblue:
Recent changes to ember-getowner-polyfill make it a true polyfill.

This lets us ship the polyfill to only environments that need it, and reduce
boilerplate code in addons.

The refactor is basically just:

```js
import { getOwner } from 'ember-getowner-polyfill';

// ....snip....
getOwner(someThing);
```

To:

```js
import Ember from 'ember';

const { getOwner } = Ember;

// ....snip....
getOwner(someThing);
```


machty:
Thanks!

PR #107, opened 2016-12-22
Re-write scheduler to be more eager / synchronous

machty:
This affords many performance wins and brings EC more in line
with async function scheduling behavior.


machty:
I've published this version to npm under the dist tag `next`. Please give this version a try and let me know if you see any regressions.

ef4:
Relevant regression: https://github.com/machty/ember-concurrency/pull/109

machty:
merged into master

PR #108, opened 2016-12-25
Implement withTestWaiter() decorator

bendemboski:
The withTestWaiter() decorator instructs a task to register a test waiting that checks if it's idle, so that asynchronous helpers in tests will wait for the task to complete.

### DIFF
--- a/API.md
+++ b/API.md
@@ -28,6 +28,7 @@ under the [Task Property API docs](TaskProperty.html).
 - [drop](TaskProperty.html#drop)
 - [keepLatest](TaskProperty.html#keepLatest)
 - [maxConcurrency](TaskProperty.html#maxConcurrency)
+- [withTestWaiter](TaskProperty.html#withTestWaiter)
 
 
 ## The full API
--- a/addon/-decorators.js
+++ b/addon/-decorators.js
@@ -1,4 +1,4 @@
-let modifierNames = ['restartable', 'drop', 'enqueue', 'maxConcurrency', 'cancelOn'];
+let modifierNames = ['restartable', 'drop', 'enqueue', 'maxConcurrency', 'cancelOn', 'withTestWaiter'];
 let decorators = {};
 
 function makeDecorator(modifierName, methodName) {
@@ -21,4 +21,5 @@ export let enqueue        = decorators.enqueue;
 export let maxConcurrency = decorators.maxConcurrency;
 export let cancelOn       = decorators.cancelOn;
 export let performOn      = decorators.performOn;
+export let withTestWaiter = decorators.withTestWaiter;
 
--- a/addon/-task-property.js
+++ b/addon/-task-property.js
@@ -138,6 +138,7 @@ export const Task = Ember.Object.extend(TaskStateMixin, {
   context: null,
   _observes: null,
   _curryArgs: null,
+  _testWaiter: null,
 
   init() {
     this._super(...arguments);
@@ -157,6 +158,11 @@ export const Task = Ember.Object.extend(TaskStateMixin, {
     }
 
     _cleanupOnDestroy(this.context, this, 'cancelAll');
+
+    if (this._testWaiter) {
+      this._registerTestWaiter();
+      _cleanupOnDestroy(this.context, this, '_unregisterTestWaiter');
+    }
   },
 
   _curry(...args) {
@@ -174,6 +180,7 @@ export const Task = Ember.Object.extend(TaskStateMixin, {
       _scheduler: this._scheduler,
       _propertyName: this._propertyName,
       _debugCallback: this._debugCallback,
+      _testWaiter: this._testWaiter,
     });
   },
 
@@ -346,6 +353,14 @@ export const Task = Ember.Object.extend(TaskStateMixin, {
   [INVOKE](...args) {
     return this.perform(...args);
   },
+
+  _registerTestWaiter() {
+    Ember.Test.registerWaiter(this, this._testWaiter);
+  },
+
+  _unregisterTestWaiter() {
+    Ember.Test.unregisterWaiter(this, this._testWaiter);
+  },
 });
 
 /**
@@ -379,6 +394,7 @@ export function TaskProperty(...decorators) {
       //_performsPath,
       _propertyName,
       _debugCallback: tp._debugCallback,
+      _testWaiter: tp._testWaiter,
     });
   });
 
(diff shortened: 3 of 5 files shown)


bendemboski:
@machty here's my stab at how this could work. I tested it in my apps, and it gets the job done! Let me know what you think.

bendemboski:
Urk, Chrome ate it

machty:
@bendemboski  This looks pretty good :)

I'm tempted to just merge this in since I haven't personally felt strongly about any particular testing enhancements for EC, and this seems at the very least like positive momentum for more creative solutions, but part of me's also concerned that it solves only a very small slice of EC testing woe, specifically tasks that wrap certain new async APIs that don't already have waiters provided by ember-testing's defaults. The biggest issues with EC testing are long timers and timer loops never settling, and I don't think this PR would help solve those problems.

I'd be curious for some other folks' feedback to see how much this might help other use cases, (and feel free to follow up if I'm missing anything), but the code changes look legit and I'm excited to see where we end up with this momentum.

bendemboski:
Yeah, I agree with you @machty. Like [this](https://github.com/machty/ember-concurrency/blob/883a071bf6eb392824fe3d4297683232b2fd7a19/tests/dummy/app/docs/testing-debugging/template.hbs#L31) issue, the problem of waiting for "non-standard" async operations is not in any way unique to ember-concurrency. However, unlike the case of long timers and timer loops, ember-concurrency provides an abstraction (i.e. tasks) that allows for an easy solution to this problem that doesn't require any knowledge of the task's implementation. So this is definitely an "opportunistic" feature, but I think a valuable one.

I think the reason this feature is possible/valuable is because there is a very clear answer to the question "how do you want test instrumentation to work?" which is "wait for the task to complete in async tests". I can't think of any similarly clear answer to that question for long timers and timer loops -- I think how you want the test instrumentation to work is highly dependent on the actual logic of your tasks, application, and tests.

I'd be very happy to help think about how ember-concurrency or Ember or some other addon could make the long timers/timer loops testing scenario easier, so feel free to pull me into any such discussions. I've spent a lot of time in the Ember testing framework, so hopefully I can be helpful :)

bendemboski:
Neither here nor there for this PR, but for anybody interested, I have been thinking about writing an addon or proposing a change to ember-test-helpers to more broadly support easy test waiting. Here are my thoughts:

```javascript
function testWaiter() {
  return false;
}

Ember.RSVP.Promise.prototype.withTestWaiter = function() {
  if (Ember.testing) {
    Ember.Test.registerWaiter(this, testWaiter);
    // There's probably a better way to do this...
    return this.finally(() => Ember.Test.unregisterWaiter(this, testWaiter));
  } else {
    return this;
  }
};
```

supports waiting for promises to resolve:

```javascript
return doSomethingThatReturnsAPromise().then(() => { /*...*/ }).withTestWaiter();
```

and

```javascript
export function withTestWaiter(cb) {
  if (Ember.testing) {
    let testWaiter = () => false;
    Ember.Test.registerWaiter(testWaiter);
    return function() {
      Ember.Test.unregisterWaiter(testWaiter);
      return cb(...arguments);
    };
  } else {
    return cb;
  }
}
```

supports waiting for callbacks to be called:

```javascript
let reader = new window.FileReader();
reader.onload = withTestWaiter((progress) => {
  // ...
});
reader.readAsText(file);
```

bendemboski:
@machty any further thinking on this? I'm finding myself wanting it in more and more of my apps & tests, and prefer not to run against a fork when I can avoid it!

machty:
@bendemboski honestly I'm leaning towards not merging without more support from other people speaking up in favor of this change. I agree that you shouldn't have to maintain a fork; maybe a reasonably compromise would be for you to maintain your test waiter hooks in your repo by reopening EC classes to support this, and if this isn't possible, I'd accept a PR for exposing useful lifecycle hooks so that it's easy to extend EC to hit use cases like test waiters.

bendemboski:
Okay, @machty I created a couple of branches (won't PR them as-is) to show some options. Either would enable me to implement the functionality that I need in my apps, or perhaps in an addon so we can see if anybody besides me starts using this.

In both branches, the unit tests are passing, the changes to `addon/` are what I would put in a PR, and a cleaned up form of the changes to `tests/dummy/` are what I would put in my apps or addon.

Aside from exporting `makeDecorator` (which I probably don't actually *need*), the only changes I needed to the core ember-concurrency code is to enable propagating the `testWaiter` information from the `TaskProperty` to the `Task` (or `Task` to `Task` when cloning) -- everything else was achievable via reopening/monkey-patching. 

https://github.com/machty/ember-concurrency/compare/master...bendemboski:custom-decorators-1 involves exposing Task factory methods that can be overridden to propagate the `testWaiter` information, and https://github.com/machty/ember-concurrency/compare/master...bendemboski:custom-decorators-2 involves exposing an `_appContext` property that will be copied around as needed, and allowing the app/addon to story any amount of data in that property, knowing that it will be copied around as needed.

What do you think? I think I kind of like the first one better since it doesn't make any assumptions about the data that might need to be propagated between objects. `Ember.copy()` would work for all the simple cases I can think of, but forcing it to happen via `Ember.copy()` rather than via whatever custom code the application wants to write seems overly restrictive, even though it simplifies the happy path a bit.

machty:
Copy/pasting from Slack: 

i had an idea

[4:53]  
re https://github.com/machty/ember-concurrency/pull/108#issuecomment-275243006

[4:53]  
lemme make a fiddle

[4:58]  
ok i can't do what i want to do because certain information is not exposed (edited)

machty [5:03 PM] 
added this Markdown (raw) snippet
​
```
TaskProperty.prototype.withCounter = function(counterKey) {
  // TODO: this.taskFn is not a real thing, but it should be
  // so that we can support this custom stubbing logic.
  
  let originalTaskFn = this.taskFn;
  
  let self = this;
  this.taskFn = function * (...args) {
    try {
      this.incrementProperty(counterKey);
      return yield * originalTaskFn.apply(self, args);
    } finally {
      this.decrementProperty(counterKey);
    }
  };
  
  return this;
}
​
export default Ember.Controller.extend({
  foo: 0,
​
  myTask: task(function * () {
    // `foo` is incremented once this task fn starts running
    // and decrements once it stops; could put test waiter logic
    // in here.
    yield timeout(1500);
  }).withCounter('foo'),
});
```

Add Comment Collapse

machty [5:03 PM] 
@bendemboski  would you be able to do everything you want to do if 1) EC maintained the import TaskProperty API and 2) exposed .taskFn ^^

machty:
Explanation of the above: we can use the `yield *` API to delegate to another fn and run it as if it's the task fn. The overriding task fn will run as a wrapper around the original taskFn (and can use getOwner or whatever hacks to determine app context or do whatever it needs to fulfill the waiter API).

bendemboski:
Support to implement this in an app or addon provided by https://github.com/machty/ember-concurrency/pull/117

PR #109, opened 2016-12-29
Regression on `next` branch

ef4:
This test passes against `master` but fails on `next`.

Now that tasks start synchronously their derived state probably needs to be updated synchronously too. Otherwise they can observe their own inconsistent state.


machty:
Thanks!

machty:
merged and fixed on master

PR #116, opened 2017-01-26
Support task decorators wrapping the task function

bendemboski:
Implement support for defining new task property modifier methods to decorate/wrap the task function itself. The `task-property-test` unit tests demonstrate how to use this to implement functionality to register a test waiter to allow tests to wait anytime certain tasks are running.

### DIFF
--- a/addon/-decorators.js
+++ b/addon/-decorators.js
@@ -1,7 +1,7 @@
 let modifierNames = ['restartable', 'drop', 'enqueue', 'maxConcurrency', 'cancelOn'];
 let decorators = {};
 
-function makeDecorator(modifierName, methodName) {
+export function makeDecorator(modifierName, methodName) {
   let fn = (...args) => {
     return (taskProperty) => taskProperty[methodName](...args);
   };
--- a/addon/-decorators.js
+++ b/addon/-decorators.js
@@ -1,7 +1,7 @@
 let modifierNames = ['restartable', 'drop', 'enqueue', 'maxConcurrency', 'cancelOn'];
 let decorators = {};
 
-export function makeDecorator(modifierName, methodName) {
+function makeDecorator(modifierName, methodName) {
   let fn = (...args) => {
     return (taskProperty) => taskProperty[methodName](...args);
   };
--- a/addon/-task-property.js
+++ b/addon/-task-property.js
@@ -365,13 +365,12 @@ export const Task = Ember.Object.extend(TaskStateMixin, {
   @class TaskProperty
 */
 export function TaskProperty(...decorators) {
-  let taskFn = decorators.pop();
   let _performsPath;
 
   let tp = this;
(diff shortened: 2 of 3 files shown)


addon/-decorators.js
```diff
@@ -1,7 +1,7 @@
 let modifierNames = ['restartable', 'drop', 'enqueue', 'maxConcurrency', 'cancelOn'];
 let decorators = {};
 
-function makeDecorator(modifierName, methodName) {
+export function makeDecorator(modifierName, methodName) {
```
machty: Why is this needed?

fwiw "decorators" (as they're called here) never really caught on and have been removed from the `next` branch. Either way you shouldn't need this for this PR.

addon/-task-property.js
```diff
@@ -382,6 +381,7 @@ export function TaskProperty(...decorators) {
     });
   });
 
+  this.taskFn = decorators.pop();
```
machty: 👍

tests/unit/task-property-test.js
```diff
@@ -0,0 +1,194 @@
...
+
+      return this;
+    }
+  },
+
+  afterEach() {
+    Ember.Test.unregisterWaiter(this.waiter);
+    delete TaskProperty.prototype.withTestWaiter;
+  }
+});
```
machty: Sorry if this is a bit nit picky but I don't really want the entire proof of concept of test waiters as an official Ember Concurrency unit test; could you slim this file down to just a simple test using the strawman example we discussed in slack? (Something that increments and decrements). We just need something that demonstrates that custom wrappers in general can be made.

bendemboski: Sure, no problem, but maybe only 20 lines of code in this test are specific to the test waiter functionality -- the majority of it is testing that things work as expected in various concurrency scenarios. I originally wrote that code because I wasn't confident in the timing of everything and wanted to prove that things worked as expected -- do you want me to axe all of that too, and basically just keep a modified form of the "withTestWaiter() decorator works" test? Or keep all of them (except the first null hypothesis one which doesn't really make sense with counters), but port them over to counter verification rather than test waiter verification?

bendemboski: I just pushed the changes that switch over to using a counter, but don't delete the concurrency scenario tests. Let me know if you want me to delete the concurrency scenario tests as well.

machty:
@bendemboski I slimmed down your PR to what I consider the essential behavior that should be tested so that we guarantee not to break the ability to extend TaskProperty's. The rest of the stuff I removed was either 1) too specific to test waiter functionality that you're trying to use this feature for or 2) fundamentally tested behavior/invariants tested elsewhere in the codebase.

machty:
Subsumed by #117

machty:
Released in 0.7.19. Thanks @bendemboski

bendemboski:
Cool, thanks for the quick turnaround!

PR #119, opened 2017-02-10
Add isSuccessful and isError to TaskInstance

mmun:
This PR adds two more derived properties to TaskInstances:

`isSuccessful`: true if the task fulfills, that is it completes without error or cancelling.
`isError`: true if the task resolves to a rejection

### DIFF
--- a/addon/-task-instance.js
+++ b/addon/-task-instance.js
@@ -109,6 +109,24 @@ let taskInstanceAttrs = {
   error: null,
 
   /**
+   * True if the task instance is fulfilled.
+   *
+   * @memberof TaskInstance
+   * @instance
+   * @readOnly
+   */
+  isSuccessful: false,
+
+  /**
+   * True if the task instance resolves to a rejection.
+   *
+   * @memberof TaskInstance
+   * @instance
+   * @readOnly
+   */
+  isError: false,
+
+  /**
    * True if the task instance was canceled before it could run to completion.
    *
    * @memberof TaskInstance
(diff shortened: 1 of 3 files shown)


addon/-task-instance.js
```diff
@@ -133,7 +151,9 @@ let taskInstanceAttrs = {
    * @instance
    * @readOnly
    */
-  isFinished: false,
+  isFinished: Ember.computed('isSuccessful', 'isError', 'isCanceled', function() {
```
machty: can't you just do `computed.or('isSuccessful', 'isError', 'isCanceled')` ?

machty: maybe i'm getting it confused w clojure

mmun: Oh, nice.

mmun: No, you're right. Didn't realize computed.or supported more than two args.

addon/-task-instance.js
```diff
@@ -365,7 +383,6 @@ let taskInstanceAttrs = {
       return;
     } else {
       if (done && value === undefined) {
-        this.set('isFinished', true);
```
mmun: This was extraneous even before this PR, as `this._finalize` will always set isFinished.

addon/-task-instance.js
```diff
@@ -235,7 +253,7 @@ let taskInstanceAttrs = {
    * @instance
    */
   cancel() {
-    if (this.isCanceled || this.isFinished) { return; }
+    if (this.get('isFinished')) { return; }
```
mmun: It wasn't possible for isCanceled to be set without isFinished being set, because the only code path that changed `isCanceled` also set `isFinished` to true.

PR #122, opened 2017-02-21
[FEATURE] [WIP] Run Loop & Ember.Evented Helpers

### LINKED ISSUE #121: [RFC] Run Loop Helpers
I think it would be super helpful so include some way to deal with Ember's run loop. There are times where you need to wait for a specific queue in the run loop to execute before you continue with the task. 

## Example 1: 

Adding data incrementally and waiting until `afterRender` to add the next set. 

```js
function * () {
  const model = this.get('model');
  const paginatedModel = paginateModel(model, 10);
  const table = this.get('table')
  
  for (let page of paginatedModel) {
    table.pushRows(page);
    yield waitForQueue('afterRender');
  }
}
```

## Example 2:

Add data to a loading modal only after it has been rendered. 

```js
function * () {
  this.setProperties({ modalLoading: true, showModal: true });
  yield waitForQueue('afterRender');
  this.setProperties({ modalLoading: false, modalData: 'foo bar' });
}
```

## Example 3:

Force the task to wait until after a specific queue has been run. This can be good if a task is called on `init` of a component but requires the dom to be ready.

```js
function * () {
  yield waitForQueue('afterRender');
  this.$().html('Hello World');
}
```

offirgolan:
Resolves #121.

Currently a work in progress with some minor blockers. Please check out the RFC (#121) for more details. 

## Todo:

- [ ]  Better assert message
- [ ]  Naming Convention
- [ ]  waitForQueue Ember < 2.8 resolution (polyfill?)
- [ ]  Better testing
- [ ]  Documentation

## Other Issues

~@machty I'm having a lot of trouble launching the dummy app / running the test cases. `ember s` goes into an infinite loop of building jsdocs.~

Changing `generateOnBuild` to `false` in ember-cli-build.js did the trick.


addon/utils.js
```diff
@@ -7,6 +7,12 @@ export function isGeneratorIterator(iter) {
           typeof iter['throw']  === 'function');
 }
 
+export function isEventedObject(c) {
+  return (c &&
+          typeof c.one === 'function' &&
+          typeof c.off === 'function');
```
machty: one interesting side effect of this is that you'll get a false positive with jQuery selectors, so you could do `yield waitForEvent($(document), 'click');`

I think this is desirable :) and is something I used in my original [EC presentation](https://github.com/machty/ec-prezo/blob/master/app/application/controller.js#L108) (though it was through a `subscribe()` interface meant for observable integration rather than a one-off "subscribe-to-a-single-event")

machty:
Some of the code you've written is reminding me that there are lot nicer patterns for setting up so-called "yieldables" in the `next` branch (intended to be the next major release of EC). I'm wondering if it wouldn't make sense to just merge `next` into master and build all new features on top of it instead; I'd definitely help show you were you'd need to move the __ec_cancel__ stuff; I promise you it's a lot nicer!

addon/utils.js
offirgolan: @machty thats actually why I did it that way 😄 . If you want me to use the `subscribe` api, I can go ahead and switch to that. I just wasnt sure it would support the functionalities we are looking for.

offirgolan:
@machty I'm up for whatever. I can also close this PR and work directly off of the `next` branch as well. Just let me know what you prefer.

offirgolan:
🔔  @machty

machty:
I'm alive! Gonna try and merge some variant of this in tonight!

PR #124, opened 2017-03-03
Make `all`, `allSettled`, `race` helpers support to cancel promises w…

Cryrivers:
…ith `__ec_cancel__`.

this would ensure these helpers to have same behaviors in normal `async` functions as in tasks.


machty:
This looks good, but I'm so very close to merging the `next` branch which already fixes this problem that I'm not sure I want to make the merge more complicated by merging this in now... so for now I'm going to keep this open and at least merge in the test case (if I haven't already written a similar test case in the `next` branch).

Cryrivers:
Understood

PR #127, opened 2017-03-13
Update to ember-cli-babel@6.

rwjblue:
This utilizes babel@6 as the underlying transpilation engine, and is compatible with babel@5 consumers (older ember-cli versions and other addons still using ember-cli-babel@5).

It is possible that there are still to be found bugs in the ember-cli-babel beta cycle, but more testers are needed to continue to flush things out.

One nice side effect of this, is that with the correct targets we end up with much less transpilation being done (and therefore the code is quite a bit more readable even without sourcemaps).

Related but distinct PR: https://github.com/machty/ember-maybe-import-regenerator/pull/4.


rwjblue:
Looks like the failure here is a chrome connect timeout (after a few of the scenarios already passed).  

@machty - Mind kicking the build?

rwjblue:
Ugh, build failed for completely WAT reasons.

PR #131, opened 2017-03-29
Adds ember-1.13 back to ember-try

MiguelMadero:
This adds back ember-1.13 support

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -2,6 +2,17 @@
 module.exports = {
   scenarios: [
     {
+      name: 'ember-1.13',
+      bower: {
+        dependencies: {
+          'ember': '~1.13.0'
+        },
+        resolutions: {
+          'ember': '~1.13.0'
+        }
+      }
+    },
+    {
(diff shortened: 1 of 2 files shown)


MiguelMadero:
The wifi here is horrible to do anything, but it looks like we're missing the `{{get}}` helper, there is a polyfill that I'll see if I can add conditionally based on the ember version once I get to a faster network

MiguelMadero:
I've been AFK but I'll have a look next week once I get back to a normal schedule.

machty:
I think you just need to add the polyfill to `devDependencies`. it's not required by EC but the docs site uses it and theres an acceptance test that does a quick check against all pages on the docs site.

MiguelMadero:
Good point. I'll give it a go on Monday and update this.

MiguelMadero:
@machty travis is happy now. It took me some time to get to this.

machty:
thanks!

MiguelMadero:
:thumbsup:

PR #145, opened 2017-04-21
Adds method to reset a tasks performCount

NLincoln:
I have no idea why `yarn.lock` changed. I can nix it if you want

### DIFF
--- a/addon/-task-state-mixin.js
+++ b/addon/-task-state-mixin.js
@@ -48,6 +48,10 @@ export default Ember.Mixin.create({
     return this._taskGroupPath && this.context.get(this._taskGroupPath);
   }),
 
+  resetPerformCount() {
+    this.set('performCount', 0);
+  },
+
   _scheduler: null,
 
 });
--- a/tests/dummy/app/docs/derived-state/template.hbs
+++ b/tests/dummy/app/docs/derived-state/template.hbs
@@ -34,7 +34,7 @@
     <code>isIdle</code>: the opposite of <code>isRunning</code>
   </li>
   <li>
-    <em><code>performCount</code></em>: The number of times the task has been performed
+    <em><code>performCount</code></em>: The number of times the task has been performed (can be reset with <code><em>resetPerformCount()</em></code>)
   </li>
   <li>
     <em><code>concurrency</code></em>: an integer representing the number of currently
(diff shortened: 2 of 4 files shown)


machty:
@NLincoln What's the use case for resetting only `performCount`? There's talk of maybe adding a `.reset()` that does a full cleanup, cancels tasks, and totally resets all properties to initial values, and I would think that `performCount` would be reset as part of that, but I'm a little hesitant to just add a method to reset only `performCount`.

NLincoln:
I think having a full reset method would be good as well, but for my use I auto refresh certain async values, and I display a loading spinner if it's the first time those tasks have run. Sometimes though I want to reset the perform count because some dependent values have changed. 

A reset method would be best, I agree. Feel free to close this.

machty:
Thanks for taking the time to work on the feature though; I could definitely use the help maintaining the lib :)

dadleyy:
fwiw I think having the ability to reset a tasks `_scheduler` would be valuable too; I've got a couple controllers with tasks defined on them that need to be cleaned up when navigating to and from the route. Right now I'm explicitly setting `task.set('lastSuccessful', null)`, which feels wrong.

NLincoln:
@dadleyy I have a few suggestions. 

Generally in our app we try to keep tasks out of our controllers for similar reasons as what you're discovering: the singleton nature of controllers makes tasks on them difficult. For that reason, maybe you could try refactoring to work with task instances? Instead of accessing `task.lastSuccessful`, you could instead set a property (e.g. `taskInstance`) on your controller somewhere with the result of calling `task.perform`.

The first thing I would do is try to move the tasks up into your route (so they can listen to `activate` and `deactivate` events), followed by setting task instances on the controller 

an example:

```js
// route.js
export default Route.extend({
  task: task(function* () { /* your task */ }),
  setupController(controller) {
    controller.set('taskInstance', this.get('task').perform());
  }
})
```

If you need to perform the tasks based on user actions, you could also perform the tasks inside of route actions.

Finally, if all that fails, you could try moving the action logic down into a component that simply yields the task and an action to perform the task. 

Without more knowledge of your use-case, I'm afraid that's all the advice I have 😅. If you'd like to talk further, I'm available as `@nlincoln` on the ember community slack.

PR #153, opened 2017-05-17
Support null/undefined values in cancelable promise helpers

bendemboski:
Similarly to how it's very nice to be able to pass a mix of promises and simple values to Ember.RSVP.{hash,all}, it's nice to be able to pass a mix of task results and simple values to ember-concurrency's hash() and all() helpers. However, if one of these values were null or undefined, the cancellation code would throw an error. This fixes that.

### DIFF
--- a/addon/-cancelable-promise-helpers.js
+++ b/addon/-cancelable-promise-helpers.js
@@ -116,10 +116,12 @@ function taskAwareVariantOf(obj, method, getItems) {
       if (hasCancelled) { return; }
       hasCancelled = true;
       items.forEach(it => {
-        if (it instanceof TaskInstance) {
-          it.cancel();
-        } else if (typeof it.__ec_cancel__ === 'function') {
-          it.__ec_cancel__();
+        if (it) {
+          if (it instanceof TaskInstance) {
+            it.cancel();
+          } else if (typeof it.__ec_cancel__ === 'function') {
+            it.__ec_cancel__();
+          }
         }
       });
     };
@@ -129,4 +131,3 @@ function taskAwareVariantOf(obj, method, getItems) {
     return promise;
   };
 }
-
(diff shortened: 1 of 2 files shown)


addon/-cancelable-promise-helpers.js
```diff
@@ -116,10 +116,12 @@ function taskAwareVariantOf(obj, method, getItems) {
       if (hasCancelled) { return; }
       hasCancelled = true;
       items.forEach(it => {
-        if (it instanceof TaskInstance) {
-          it.cancel();
-        } else if (typeof it.__ec_cancel__ === 'function') {
-          it.__ec_cancel__();
+        if (it !== null && it !== undefined) {
```
bendemboski: I decided to check this way rather than just a truthy check for a pedantic and maybe-silly reason. Theoretically somebody could provide a non-`null` non-`undefined` falsey value with an `__ec_cancel__` property on it. I can't imagine why somebody would want to, but I also can't see a reason why we shouldn't call such a method, so I made the check stricter.

tests/unit/cancelable-promise-helpers-test.js
```diff
@@ -307,6 +307,31 @@ test("hash cancels children if parent is canceled", function(assert) {
   assert.equal(obj.get('child.concurrency'), 0);
 });
 
+test("yieldable helpers work with null/undefined values", function(assert) {
```
bendemboski: This only tests the `hash` helper, but the code path is common across all helpers. I could write tests for the other helpers, but it seemed unnecessary -- let me know if you think otherwise.

machty: No just the one test is fine.

addon/-cancelable-promise-helpers.js
machty: I'd still prefer a simply truthy check here

bendemboski: Okay, fixed and rebased

machty:
Thanks!

PR #154, opened 2017-05-18
`enqueueWithPriority` task modifier

vitch:
This modifier would allow you to pass a `sortFunc` which is used to
sort the queue directly before taking any tasks off it. This means that
you can add a bunch of tasks to a queue but mark some as more important
than others.

The contents of this commit are just a spike based of some advice from
@machty in slack. If this general approach makes sense then I could
try to make sense out of the tests and round this out into a more
complete PR...

### DIFF
--- a/addon/-buffer-policy.js
+++ b/addon/-buffer-policy.js
@@ -24,7 +24,7 @@ export const enqueueTasksPolicy = {
   }
 };
 
-export const enqueuePriorityPolicy = {
+export const enqueueWithPriorityPolicy = {
   requiresUnboundedConcurrency: true,
   schedule(scheduler) {
     scheduler.queuedTaskInstances.sort(this.sortFunc);
--- a/addon/-buffer-policy.js
+++ b/addon/-buffer-policy.js
@@ -24,6 +24,17 @@ export const enqueueTasksPolicy = {
   }
 };
 
+export const enqueuePriorityPolicy = {
+  requiresUnboundedConcurrency: true,
+  schedule(scheduler) {
+    scheduler.queuedTaskInstances.sort(this.sortFunc);
+    saturateActiveQueue(scheduler);
+  },
+  getNextPerformStatus(scheduler) {
+    return numPerformSlots(scheduler) > 0 ? 'succeed' : 'enqueue';
+  }
+};
+
 export const dropQueuedTasksPolicy = {
   cancelReason: `it belongs to a 'drop' Task that was already running`,
   schedule(scheduler) {
--- a/addon/-property-modifiers-mixin.js
+++ b/addon/-property-modifiers-mixin.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import Scheduler from './-scheduler';
 import {
   enqueueTasksPolicy,
-  enqueuePriorityPolicy,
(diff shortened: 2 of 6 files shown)


vitch:
I have used this branch in our app and it works beautifully... I use `enqueuePriority` on a `taskGroup` and it "just works" and the API feels nice and clean.

I also cleaned up the demo a bit and added API documentation.

A few outstanding points:

 1. I'm not completely sold on the name. How about `enqueuePrioritised` or `enqueueWithPriority`?
 2. I'd love to add a test but am having trouble figuring out what it would look like. I guess it would live in `buffering-test`? Can you give me any pointers on how you would approach it?

machty:
How about `enqueueWithPriority` to avoid the American/British English debate over "ize" and "ise". 

buffering-test makes more sense for the other buffer policies but I think this is different enough to warrant its own file, e.g. `tests/priority-queue-test.js` or something. You could write a test that looked similar to some of the tests in [task-test.js](https://github.com/machty/ember-concurrency/blob/master/tests/unit/task-test.js). I find it's easy to use a pattern of pausing tasks on defer promises that you can manually resolve whenever you're ready to proceed with the next assertion, like in [this example](https://github.com/machty/ember-concurrency/blob/master/tests/unit/task-instance-test.js#L155).

vitch:
> How about enqueueWithPriority to avoid the American/British English debate over "ize" and "ise".

I'm didn't even realize (:p) that I spelled it the British way! I've changed it to `enqueueWithPriority` but I'd also be happy with `enqueuePrioritized` if you prefer that!

I'll look into the test this afternoon and add something to the PR...

vitch:
Added some test coverage and rebased on top of latest `master`.

I don't *think* that the tests really need to be too concerned with asynchronicity since that is tested elsewhere. So I just verified that the order worked as expected.

There is an interesting side effect of the way it's built  is that the first `task` `perform`ed in each runloop doesn't respect the sorting - the sort function is only called after a task completes... I'm not sure if this is a problem or not?

machty:
@vitch I'm not sure I understand this side effect you're describing... could you maybe provide an example?

vitch:
[This assert](https://github.com/machty/ember-concurrency/pull/154/files#diff-6ce76e1ae54155c2caf11b1f4737ac73R89) kind of sums it up - it would be nice if that was `['f', 'c', 'b', 'e', 'd', 'a']` or if [this assert](https://github.com/machty/ember-concurrency/pull/154/files#diff-6ce76e1ae54155c2caf11b1f4737ac73R30) could be `['f', 'e', 'd', 'c', 'b', 'a']`.

vitch:
I put together a reproduction of our use-case for `enqueueWithPriority` - you can see it here:

https://github.com/vitch/ec-prioritised
https://vitch.github.io/ec-prioritised/

In [this additional code](https://github.com/vitch/ember-concurrency/issues/1) you can another change I had to introduce for our particular instance.

Let me know what you think of these and what the best way forward is with this stuff? Should it go into `ember-concurrency` or should I start packaging it as a companion addon?

machty:
Thank you for putting this together so many years ago. A lot has changed and I think it's best to close at this point but if you're interested in reviving this, let's discuss in the `#e-concurrency` Discord channel.

PR #156, opened 2017-06-06
Refactor to use async/await ember-native-dom-helpers in testing

cibernox:
This only affects acceptance and integration tests, not the tests involving your special test utilities with the same `click`/`find`/etc... names.

My original intention was to remove jQuery from the addon entirely, but then I realised that your test helpers use it, so I left it in a simple refactor to use ember-native-dom-helpers and async await in the tests that are not specifically using your helpers.


machty:
Thank you for this!

Yeah, I'm gonna try removing (maybe extracting?) the code I wrote (it was initial meant to make the case for the capybara-style assertions I was proposing for testing unification).

cibernox:
the failure seems spureous. Can you rebuild?

cibernox:
🍏

machty:
thankya!

Issue #162, opened 2017-06-30
Allow concurrency that depends on arguments passed to a task


orf:
Scenario: 

You have an `editor` component that periodically auto-saves content using a concurrent task. If the task lives on the component then switching away from the page could loose saves.

So you move the task to a service. However this service is not very useful, as if you use something like `keepLatest` then only one save can run at a single time, and saves for other models can be lost. What you really want is a `keepLatest` that depends on an argument that is passed to the task.

I.E, if your first parameter to the task is `model`, you want the `keepLatest` functionality *for each task launched with the same model instance*. So you could have two concurrent tasks with different models running, but repeated calls to that task with the same model would trigger a queue or drop if one is already running.

I hope this makes sense, sorry if it doesn't! I was looking at the source code, couldn't it be possible to pass in the task instance or the arguments passed to the tasks `buffer policy` as a start, then maybe have a policy that accepts a function as it's first argument, returning a stable `key` from the given task arguments. Then the policy can apply the current logic to tasks that share that unique key?

rlivsey:
I've been experimenting with what I've been (badly) calling "contextual services" in [ember-contextual-services](https://github.com/minutebase/ember-contextual-services) to solve this in my app.

Basically it gives you services for your model instances.

Eg (from the readme):

```js
//app/contextual-services/person.js
import ContextualService from 'ember-contextual-services';
import { task, timeout } from 'ember-concurrency';

export default ContextualService.extend({
  debouncedSave: task(function * () {
    yield timeout(1000)
    yield this.get('model').save();
  }).restartable();
});
```

gabrielgrant:
Perhaps this is a silly question, but if the task is something performed on the model, is there a reason the task shouldn't live on the model?

c-emil:
@gabrielgrant That's actually not a silly question at all! In fact, that's a very good idea that solves the problem of different models for the same task. Thanks a lot for the tip.

gabrielgrant:
@c-emil I think one potential downside of this is that if the task is doing something fairly specific to a particular component (or other part of your app), it could create a coupling in the "wrong" direction -- my understanding is that generally Ember best practice is to keep the data layer fairly functionality-agnostic, and contain that functionality instead within controllers and/or components.

In the case the task is something more generally applicable, though, yes, it does seem model-tasks could be useful

c-emil:
[...]

PR #163, opened 2017-07-05
WIP: Add support for multiple task channels

orf:
Ref #162 

This is an initial stab at implementing multiple 'channels' per given task. It's no where near done, but I figured I would try and see if it was possible - and it seems so. The code is a super proof of concept and is a bit nasty (i'm looking at you `taskInstancesByChannel`).

Each task has a 'channel', which is just a string key returned from a user-defined function based on the tasks arguments, e.g:

```js
thing: task(function*(name){ 
      yield timeout(100); 
      console.log(name);
}).channel(name=>name).drop()
```

Each time `thing` is performed the arguments would be passed to the function given to channel, which must return a string. The buffer policies for the task perform the same, but only apply **to the subset of tasks with the same channel identifier**.

So, in theory, the `thing` task could have two concurrently executing tasks with different channel identifiers, `tom` and `henry`. If another task is performed (e.g `thing.perform('tom')`), then the `drop` would kick in and the task would not execute. However `thing.perform('james')` would execute concurrently with `tom` and `henry`.

I implemented this by separating the buffer policies from the scheduler. Instead of passing in the scheduler instance to the buffer policies we pass in arrays of queued/active task instances, which allows the scheduler to pass a given subset in.

Please tear this merge request apart and let me know if it's worth exploring in more detail, if so I will invest some more time into it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ dist: trusty
 
 cache:
   yarn: true
+  directories:
+      - $HOME/.cache/bower
 
 addons:
   apt:
--- a/addon/-buffer-policy.js
+++ b/addon/-buffer-policy.js
@@ -1,23 +1,25 @@
-const saturateActiveQueue = (scheduler) => {
-  while (scheduler.activeTaskInstances.length < scheduler.maxConcurrency) {
-    let taskInstance = scheduler.queuedTaskInstances.shift();
+import {spliceTaskInstances} from './-scheduler';
+
+const saturateActiveQueue = (maxConcurrency, queuedTaskInstances, activeTaskInstances) => {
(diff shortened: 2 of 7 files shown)


orf:
Tests are failing on ember-beta, but master does as well on my machine, FYI.

machty:
I'm really sorry for not following up with this; I don't know whether I saw it and forgot or just missed it due to prior bad inbox habits.

Task channels seems like a cool idea that would compose well with other features that people have requested. I'm uncertain how much of this PR I can use but I'm leaning toward the idea that every task internally have multiple channels, and by default their channel mapper fn just returns the same static string.

c-emil:
Any chances this will be implemented in near future? Or any ideas how for workarounds until it's implemented?

PR #167, opened 2017-07-08
Allow tasks to be specified as linked or unlinked

### LINKED ISSUE #161: The sinister self-cascade-cancel footgun / megatroll
This issue has come up enough times that I figured I'd make an official issue about it so I could amass ideas:

When you have a Component task that performs and yields a task on a Service, and the Service changes the state in such a way that it causes the Component to unrender, then componentTask will be canceled and cascade-cancel serviceTask.

I personally run into this issue all the time because in my mobile app, we use tasks on services to coordinate global state changes to the app, e.g. a button on one page will run a component task, which runs a service task to cause a page transition. If I forgot to "sever" the entanglement between componentTask and serviceTask, then the component gets destroyed, canceling componentTask and serviceTask, and hence any code that is suppose to run on serviceTask after the page transition / re-render will not run.

My hackish solution so far has been to introduce the following util macro in our code base to intentionally stick out like a sore thumb and clarify what's happening:

```
/*
 * ember-concurrency cancels tasks (and recursively cancels
 * any yielded child tasks) when the object a task lives on
 * is destroyed. This is a good default, but is surprising
 * in cases where, say, a component yields a task.perform on
 * a service, and then that component gets unrendered as part
 * of what the service task does. So by wrapping the serviceTask.perform
 * in `goOnWithoutMe`, the component task will still be canceled
 * but the service task will go on even though the component
 * task that called it was canceled.
 **/
export function goOnWithoutMe(taskInstance) {
  return Ember.RSVP.resolve(taskInstance);
}
```

which would be used as follows:

```
componentTask: task(function * () {
  yield goOnWithoutMe(this.get('myService.someTask').perform())
})
```

It already sucks that this hack is required to facilitate an otherwise nice pattern, but there are even more problems:

- it's really hard to explain when and why you'd need this hack; when this gotcha bites, things just seem to stop working part way through with no messaging or obvious explanation as to why
- while the promise-wrapping fixes the problem of undesirable-implicit-cancels-upon-destruction, it means explicit cancels that you _would_ expect to cancel `serviceTask` will not actually cancel `serviceTask` since promises aren't cancelable.

So, we need a solution to this problem. I've proposed an `.atomic()` task modifier in other related issues; you specify a serviceTask as `atomic()` which would prevent it from canceling partway through, but again, you'd have to know in advance that you need this modifier, and if you forget it, you get the Silent Treatment.

machty:
Fixes #161 

This PR detects the dreaded ["self-cancel loop/footgun"](https://github.com/machty/ember-concurrency/issues/161) and `console.warn`s that you need to use `someTask.linked().perform()` or `someTask.unlinked().perform()` to annotate how you want it to behave.

Aside from annotating the desired behavior in the case of the above footgun, `.linked()` doesn't actually do anything different than a normal `yield someTask.perform()`, and will warn you if you try and use it without immediately returning/yielding.

But `.unlinked()` on the other hand has use cases beyond fixing self-cancel loops. Semantically, the following are equivalent:

```
yield this.get('aTask').unlinked().perform()
yield Promise.resolve(this.get('aTask').perform());
```

Which is to say: "wait for this task instance to run to completion, but if I, the parent task, get canceled, do NOT cancel this child task"; previously the only way to do this want to cast to a promise, but this was hacky and unclear and it's good to have some standardized terminology as to what's happening.

### DIFF
--- a/addon/-helpers.js
+++ b/addon/-helpers.js
@@ -1,15 +1,14 @@
 import { get } from '@ember/object';
 import { assert } from '@ember/debug';
 import { bind } from '@ember/runloop';
-import { Task } from 'ember-concurrency/-task-property';
 
-export function taskHelperClosure(taskMethod, _args, hash) {
+export function taskHelperClosure(helperName, taskMethod, _args, hash) {
   let task = _args[0];
   let outerArgs = _args.slice(1);
 
   return bind(null, function(...innerArgs) {
-    if (!(task instanceof Task)) {
-      assert(`The first argument passed to the \`perform\` helper should be a Task object (without quotes); you passed ${task}`, false);
+    if (!task || typeof task[taskMethod] !== 'function') {
+      assert(`The first argument passed to the \`${helperName}\` helper should be a Task object (without quotes); you passed ${task}`, false);
       return;
     }
 
--- a/addon/-task-instance.js
+++ b/addon/-task-instance.js
@@ -25,6 +25,16 @@ const GENERATOR_STATE_HAS_MORE_VALUES = "HAS_MORE_VALUES";
 const GENERATOR_STATE_DONE = "DONE";
 const GENERATOR_STATE_ERRORED = "ERRORED";
 
+export const PERFORM_TYPE_DEFAULT  = "PERFORM_TYPE_DEFAULT";
+export const PERFORM_TYPE_UNLINKED = "PERFORM_TYPE_UNLINKED";
+export const PERFORM_TYPE_LINKED   = "PERFORM_TYPE_LINKED";
+
+let TASK_INSTANCE_STACK = [];
+
+export function getRunningInstance() {
+  return TASK_INSTANCE_STACK[TASK_INSTANCE_STACK.length - 1];
+}
+
 function handleYieldedUnknownThenable(thenable, taskInstance, resumeIndex) {
   thenable.then(value => {
     taskInstance.proceed(resumeIndex, YIELDABLE_CONTINUE, value);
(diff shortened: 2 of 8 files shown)


Turbo87:
@machty is there a way to use `unlinked` from within templates? something like `{{perform someTask unlinked=true}}` or `{{unlinked-perform someTask}}`?

machty:
@Turbo87 No, but sounds like we need something like that. Not sure if it should be `unlinked=true` but keep in mind that to appease the "self cancel loop" message, you sometimes also have to explicitly mark the task as linked. Part of me wishes I'd simply made `task.linked` and `task.unlinked` be ES5 getters rather than function calls... I don't remember why I didn't do it that way, just that maybe it seemed too magical? But it would have solved this problem in both JS and template world.

Turbo87:
yeah, might be useful. I've solved it for now with a custom `unlinked-perform` helper :)

machty:
@Turbo87 The real insight here is that e-c's got the linking model slightly wrong; currently, a "parent" task and "child" TaskInstance are linked when one `yield`s another; instead the linking should be determined when one _performs_ the other; in the absence of this there are some cases where you want a task to be linked regardless of whether you await/yield it. 

That said, this all gets very confusing when a component is declaring tasks and passing it down to child components... in this case, who is the "parent"?

Turbo87:
I totally agree that this is confusing. In our specific case it was like this:

the controller had an `authenticateTask` which it passed to the `login-form` component via `authenticate=(perform authenticateTask)`. that component had a `submitTask` which called `yield this.authenticate()` when the submit button was clicked.

PR #170, opened 2017-07-28
Use "New Module Imports"

Turbo87:
see https://github.com/ember-cli/ember-rfc176-data

### DIFF
--- a/addon/-cancelable-promise-helpers.js
+++ b/addon/-cancelable-promise-helpers.js
@@ -1,10 +1,8 @@
-import Ember from 'ember';
+import RSVP, { Promise } from 'rsvp';
 import TaskInstance from './-task-instance';
 import { yieldableSymbol } from './utils';
 
-const RSVP = Ember.RSVP;
-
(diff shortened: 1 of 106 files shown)


Turbo87:
CI failure seems unrelated 😞

machty:
Thanks! How does this work for older versions of Ember that want to use e-c?

Turbo87:
`ember-cli-babel` will essentially just translate those imports into `const foo = Ember.foo`. Once Ember actually supports those import paths we'll use version detection in `ember-cli-babel` to disable that transform.

tl;dr it should just work 😉

machty:
Does it require that all consumers of the new e-c also use babel 6.6.0? Or do addons get compiled with their preferred version of ember-cli-babel?

Turbo87:
> do addons get compiled with their preferred version of ember-cli-babel

yep, that's how it works

Turbo87:
@machty unfortunately I am very stupid and introduced a bug in this PR. My statement above about babel version used for the addon transpilation does not apply to the app folder but I've converted the files in there too. I'll fix that asap tomorrow. Sorry! 😞

PR #187, opened 2017-11-14
feat(wait-for/event): support native DOM EventTargets

### LINKED ISSUE #180: waitForEvent: support native DOM events
`waitForEvent` currently does not support native DOM events.

```js
const image = new Image();
next(() => { // make sure the load event fires after the handler is registered
  image.src = 'rick-astley.gif';
});
yield waitForEvent(image, 'load');
```

```
Error: Assertion Failed: [object HTMLImageElement] must include Ember.Evented
  (or support `.on()`, `.one()`, and `.off()`) to be able to use `waitForEvent`
```

For the time being you can wrap the element in a jQuery object, like so:

```js
const image = new Image();
next(() => { // make sure the load event fires after the handler is registered
  image.src = 'rick-astley.gif';
});
yield waitForEvent($(image), 'load');
```

buschtoens:
Closes #180.

Also removes `.on()` from the assertion message, as it was not supported.

### DIFF
--- a/addon/-wait-for.js
+++ b/addon/-wait-for.js
@@ -27,13 +27,32 @@ class WaitForEventYieldable {
   }
 
   [yieldableSymbol](taskInstance, resumeIndex) {
+    let unbind = () => {};
     let fn = (event) => {
+      unbind();
       taskInstance.proceed(resumeIndex, YIELDABLE_CONTINUE, event);
     };
-    this.object.one(this.eventName, fn);
-    return () => {
-      this.object.off(this.eventName, fn);
-    };
+
+    if (typeof this.object.addEventListener === 'function') {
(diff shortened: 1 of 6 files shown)


machty:
I'm sorry it's taken me so long to look at this; I think this looks pretty good :).

Do you still feel that this is a valuable addition to e-c? Could I convince you to add one more commit that adds docs to [this page](http://ember-concurrency.com/#/docs/events)?

buschtoens:
No problem and thank you.

Yes, I do believe this is still a valuable addition, especially since Ember Core is going to ~deprecate jQuery~ *make jQuery optional*: [RFC 294](https://github.com/emberjs/rfcs/pull/294)

I'll add another commit that documents this feature. :+1:

PR #189, opened 2017-11-19
Fix for "Cancelling task does not update group's isRunning property"

### LINKED ISSUE #172: Cancelling task do not update group's isRunning property
*version*: 0.8.7
*Steps to reproduce*:
  * Create a long-running task and assign it to group
  * Perform task
  * Cancel task

*Expected result*: Group's `isRunning` property is `false` after cancellation
*Actual result*: Group's `isRunning` property is `true`

GendelfLugansk:
This commit fixes #172


machty:
We shouldn't be using `next`, as this is going to 1. likely slow things done by kicking continuation of the task to a separate event loop via setTimeout(0) and 2. break a lot of people's tests that rely on the established timing of tasks. I'm not sure I know the alternative fix at the moment, but if you've still got availability, could you at least provide a breaking test that I or someone else might be able to fix the bug for?

GendelfLugansk:
Added test in another PR

PR #194, opened 2018-01-09
upgrade to 2.18

snewcomer:
wip upgrade to 2.18
 
Still want to support back to 1.13?

- [x] ember-modules-codemod
- [x] ember-qunit-codemod
- [x] fix eslint errs (unused vars)
- [x] ember-try config > 2.4
- [x] loading-spinner (crap) component instead of ember-font-awesome if you want.  Can easily revert.  Upgraded versions of ember-font-awesome require lots of ember-try configs.
- [ ] update ember-try config to run w/ 2.16


machty:
I'm told that the convention is to support at least the last two LTS versions 2.12 and 2.16. Do you know anything in particular that will break 1.13 support?

tests/dummy/app/index.html
```diff
@@ -3,23 +3,23 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>ember-concurrency</title>
```
alexander-alvarez: ^

machty:
I'm not sure which dependency changed to break this, but basically [this code](https://github.com/machty/ember-concurrency/blob/master/tests/helpers/generator-tests.js#L16) is trying to call the `wait()` testing helper to wait for the world to settle. Maybe it's moved elsewhere? Could you try putting a debugger before that line and just exploring if there's any other way to access `wait()` from that context?

snewcomer:
@machty yeah I'm seeing diff behaviors whether I use `app.testWaiters.wait` or `wait`/`settled` from `@ember/test-helpers`.  These are just internal test helpers, right?  What if we just did:

```
yield settled();
....then do find of the `selector`
```

For instance, it seems like `options` hash to `find` (`options.timeout` and  `let count = typeof options.count...` is only used/tested in the `test-utilities-test`.

machty:
@snewcomer what do you mean different behaviors? Seems like `app.testWaiters.wait` is straight up undefined?

snewcomer:
@machty Hm yeah you are right.  Removed that code [here](https://github.com/machty/ember-concurrency/pull/194/commits/7ccc2ca1a82fbfc1806b3c4dcc01c1271dc8a797) and still got [this](https://travis-ci.org/machty/ember-concurrency/builds/330090204?utm_source=github_status&utm_medium=notification) test failure, but only in one of the travis builds


https://travis-ci.org/machty/ember-concurrency/jobs/330090209

snewcomer:
@machty ok so I upgraded the testing structure to the new qunit nested syntax.  Everything seems a-ok but for 2 tests - for ex// `ajax-throttling` example that have `while` loops in their route or component hooks on init.

https://github.com/machty/ember-concurrency/blob/master/tests/dummy/app/components/ajax-throttling-example/component.js#L6

Specifically, this [line](https://github.com/emberjs/ember-test-helpers/blob/c27850b0d67ab709e13305268560aefb569e8d84/addon-test-support/%40ember/test-helpers/wait-until.js#L47) from ember-test-helpers seems to be thrown in a while loop.

snewcomer:
Further looking into the infinity loop problem, `run.hasScheduledTimers()` returns true, which I guess makes sense.

https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/settled.js#L146

machty:
@snewcomer I don't want to keep sending you on a wild goose chase here and I think you might be missing some context: with some of these testing APIs I was exploring an alternative way of testing Ember apps that would avoid a lot of issues with slowdowns / halted tests due to long timers or timer loops. By using so called "eager assertions", these new kinds of tests would wait until some assertion criteria was met, and then proceed onward, as opposed to the classic Ember testing model where you're constantly waiting for the world to "settle" (including outstanding timers) before synchronously running assertions against the DOM.

Ultimately, I didn't succeed in convincing people this was the better way to go (and honestly I'm not sure I've convinced myself either, but it was worth exploring), and as you mention in the linked issue, ember-lifeline is an alternative to expressing while loops that sidesteps a lot of these issues, but that said, I don't want to bring in lifeline into the dummy app just yet until we e-c / e-l maintainers have decided that we ought to converge on some patterns somehow.

How bout I make some time this weekend to finish the work you've started? At the moment I can't fully dig in to explore the issue further but I don't want you to feel like you're spinning your wheels.

snewcomer:
That sounds great!  Lmk how I can help/review.

machty:
Closing in favor of now-merged #199

Thank you so much for the hard work!

PR #199, opened 2018-01-21
upgrade to 2.18

machty:
Rebased/merged version of #194


machty:
@snewcomer sorry I might have overcomplicated things by creating this separate PR, not sure; I ended up squashing / rebasing things off of master, now I'm wondering what else I need to review here. I see two `Ember.testing` checks that I want to get rid of (particularly since they'll show up in the documentation sites since they're in snippet blocks). Is there anything else?

machty:
@snewcomer Got things passing; root issue seems to be at some point ember-testing changed to await settling at the end of a test (even if the test itself doesn't care about settling). So I just added a `cancelTimers()` at the end of our internal `test` function.

Does this look ready to merge to you?

snewcomer:
@machty Ah great idea!  As long as you are ok with the loading component, yeah it looks great.

Issue #200, opened 2018-01-24
Replace all `/app` code with re-exports from `/addon`


machty:
Instead of having helpers/initializers directly implemented via files in `/app`, all of their code should move to `/addon`, and the files in `/app` should simply do `export from` statements ([example](and we shouldn't shy away from these ergo improvements)).

This is a nice convention that most addons adopt that makes it easier to override/extend addon behavior, but would also let us avoid the problem where the `app/` tree can't adopt the new `@ember` import conventions without breaking people's builds due to older versions of ember-cli.

corrspt:
Looking at the merged PR, should this issue be closed?

machty:
Yup :)

PR #201, opened 2018-01-25
move app folder to addon folder and add reexports

spencer516:
For: https://github.com/machty/ember-concurrency/issues/200

### DIFF
--- a/addon/helpers/cancel-all.js
+++ b/addon/helpers/cancel-all.js
@@ -1,0 +1,17 @@
+import Ember from 'ember';
+import { taskHelperClosure } from 'ember-concurrency/-helpers';
+
+const CANCEL_REASON = "the 'cancel-all' template helper was invoked";
+
+export function cancelHelper(args) {
+  let cancelable = args[0];
+  if (!cancelable || typeof cancelable.cancelAll !== 'function') {
+    Ember.assert(`The first argument passed to the \`cancel-all\` helper should be a Task or TaskGroup (without quotes); you passed ${cancelable}`, false);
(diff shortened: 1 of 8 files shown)


addon/helpers/task.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
```
alexander-alvarez: `import { helper } from '@ember/component/helper';`

addon/helpers/perform.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
```
alexander-alvarez: `import { helper } from '@ember/component/helper';`

addon/helpers/cancel-all.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
```
alexander-alvarez: `import { helper } from '@ember/component/helper';`

spencer516:
Happy to make that change — but to be sure — using the new module imports was [recently reverted](https://github.com/machty/ember-concurrency/commit/1dd1600a67f6d906ee327347b9dd8251a95e3810) as it caused some issues with older versions of ember. (This PR, of course, should make that no longer an issue)...but just want to be sure that change shouldn't be a separate PR.

machty:
@spencer516 the new import style is safe so long as it's in the `addon/` folder (which it is). The pattern of keeping everything in `addon` and just re-exporting things in `app` facilitates this.

machty:
Thank you!

PR #204, opened 2018-02-12
Fastboot rendering for docs

ef4:
I noticed that deep-linking is broken on the docs site (try clicking http://ember-concurrency.com/docs/introduction).

This PR fixes that by pre-rendering all the docs pages in fastboot at deploy time. You can see a sample deployment at https://ef4.github.io/ember-concurrency/


lib/prember-urls.js
```diff
@@ -0,0 +1,17 @@
+const jsdom = require("jsdom");
+const { JSDOM } = jsdom;
+
+module.exports = function urls(distDir, visit) {
+  return visit('/docs/introduction').then(page => {
+    return page.html();
+  }).then(html => {
+    var dom = new JSDOM(html);
+    var urls = ['/'];
+    for (let aTag of [...dom.window.document.querySelectorAll('.side-menu a')]) {
```
ef4: This is where I'm deciding which links to follow when pre-rendering.

package.json
```diff
@@ -49,7 +50,9 @@
     "ember-notify": "~5.2.1",
     "ember-resolver": "^4.0.0",
     "ember-source": "~2.18.0",
-    "loader.js": "^4.2.3"
+    "jsdom": "^11.6.2",
+    "loader.js": "^4.2.3",
+    "prember": "0.3.0-alpha.3"
```
ef4: This is an alpha because it depends on unreleased PRs in ember-cli-fastboot and broccoli-asset-rev. I won't be mad if you don't want to merge until these are real releases.

tests/dummy/app/components/events-example/component.js
```diff
@@ -1,9 +1,9 @@
 import Evented from '@ember/object/evented';
-import Controller from '@ember/controller';
+import Component from '@ember/component';
```
ef4: I made this into a component and used `on('didInsertElement')` instead of `on('init')` in order to avoid cluttering up your example snippts with isFastBoot conditionals.

(Touching jQuery on init in fastboot blows up)

lib/prember-urls.js
ef4: (Anything we don't pre-render will still render normally, due to having a 404.html file. The only downside of that is that the status code is 404 instead of 200.)

machty:
Thanks @ef4, you're the best!

PR #212, opened 2018-03-10
Add a testimonials page

sukima:
This is a page that will list user testimonials. The intent is to offer collaboration with the community by giving them a platform to posts stories where ember-concurrency has really helped.

To accomplish both the presentation and the ability for users to contribute through Pull Requests this uses a [3rd party API](https://github.com/sukima/ember-concurrency-testimonials-api/tree/master#how-to-contribute). This API is simply a GitHub page with static JSON files. They are created by parsing Markdown documents. This is possible because GitHub adds CORS headers for files served through GitHub Pages.

A user will contribute by creating a Markdown document (with some YAML front matter) like they would with a static Jekyll site. The build tools will convert them to JSON files and push to `gh-pages`. The static files act as a live API which this feature uses to populate its model(s).

Currently the API has a bunch of dummy data. This will change when this is released. The feature can handle an empty result set with a _Be the first_ message.

## Requirements Needed

- [ ] Who should own the static API repository?
- [ ] The API has knowledge of the number of total testimonials. Do we want to show these numbers in the template?
- [ ] Do we want to pre-fill some testimonials before initial launch?
- [ ] When a user clicks create new they are redirected to the API repo README; how do we want to word the contributions instructions?

## Demo

https://sandbox.tritarget.org/docs/testimonials

ec-testimonials

### DIFF
--- a/tests/dummy/app/components/testimonial-box/component.js
+++ b/tests/dummy/app/components/testimonial-box/component.js
@@ -1,0 +1,11 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  classNames: ['testimonial-box'],
+  attributeBindings: ['role'],
+  role: 'button',
+
+  click() {
+    return this.get('onClick')(this.get('testimonial'));
+  }
+});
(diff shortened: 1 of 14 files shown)


machty:
Thanks for putting this together; doing some housekeeping and I think it's best to close this for now.

PR #213, opened 2018-03-15
Introduce tagged timers to help test timer loops and long timers

machty:
Example: the following snippet has a timer loop that will
indefinitely pause your test suite (because ember-testing
internally tries to wait for all timers to elapse before
considering the app as "settled"):

```
  loopingTask: task(function * () {
    while(true) {
      // doStuff()
      yield timeout(200);
    }
  }),
```

To aid in these cases (and other cases with long timers) you
can now pass a second param into `timeout()` which is a test tag:

```
  numIterations: 0,
  loopingTask: task(function * () {
    while(true) {
      this.incrementProperty('numIterations');
      yield timeout(200, 'server-poll-timer');
    }
  }),
```

When testing, a `timeout()` with a test tag will indefinitely
pause the task at the `yield timeout(...)` without causing
ember-testing to stop execution of your test. From within
a test, you can use the new `resumeTimers()` utility method
to resume any timers with a matching test tag. `resumeTimers(testTag)`
returns the number of matching timers that were resumed (so you
can assert that the task was paused at the timer where you expected).

Timer tags allow you to step through timer loops one iteration at
a time without indefinitely pausing or even slowing down your test suite:

```
  test('polling loop works', async function(assert) {
    await visit('/route-with-timer-loop');

    assert.equal($('.num-iterations').text(), '1');
    assert.equal(await resumeTimers('server-poll-timer'), 1);

    assert.equal($('.num-iterations').text(), '2');
    assert.equal(await resumeTimers('server-poll-timer'), 1);

    assert.equal($('.num-iterations').text(), '3');
  });
```

### DIFF
--- a/addon/testing.js
+++ b/addon/testing.js
@@ -1,0 +1,73 @@
+import {get} from '@ember/object';
+import {yieldableSymbol, YIELDABLE_CONTINUE} from './yieldables';
+import {TimeoutPromise} from './-timeout';
+
+let TEST_TIMERS = {};
+let TIMES_SINCE_LAST_CLEAN = 0;
+
+function cleanupTestTimers() {
+  let newTestTimers = {};
+  for (let k in TEST_TIMERS) {
+    let timers = TEST_TIMERS[k];
+
+    newTestTimers[k] = [];
+
+    timers.forEach(timer => {
+      // assume task instances from previous test runs have been properly canceled
+      if (get(timer.taskInstance, 'isRunning')) {
+        newTestTimers[k].pushk(timer);
+      }
+    });
+  }
+  TEST_TIMERS = newTestTimers;
+}
+
+export class TestTimeoutPromise extends TimeoutPromise {
+  constructor(ms, label) {
+    super(ms);
+    this.label = label;
+  }
+
+  [yieldableSymbol](taskInstance, resumeIndex) {
+    if (TIMES_SINCE_LAST_CLEAN++ > 20) {
+      TIMES_SINCE_LAST_CLEAN = 0;
+      cleanupTestTimers();
+    }
+
+    let intervalId;
+    let label = this.label;
+    TEST_TIMERS[label] = TEST_TIMERS[label] || [];
+    TEST_TIMERS[label].push({
+      label,
+      taskInstance,
+      hasResumed: false,
+      resume() {
+        if (this.hasResumed) {
+          return;
+        }
+        this.hasResumed = true;
+        taskInstance.proceed(resumeIndex, YIELDABLE_CONTINUE, null);
+      },
+    });
+
+    return () => {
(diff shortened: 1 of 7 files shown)


machty:
This is still pretty messy and broken atm; in particular, just realized that every `resumeTimers` should be `await`ed... not sure how/why it was passing locally.

addon/testing.js
```diff
@@ -0,0 +1,73 @@
+import {get} from '@ember/object';
+import {yieldableSymbol, YIELDABLE_CONTINUE} from './yieldables';
+import {TimeoutPromise} from './-timeout';
+
+let TEST_TIMERS = {};
```
bendemboski: FYI, when using the new test API, you can `import { getContext } from '@ember/test-helpers'` to get the current test context object (if a test is running), i.e. `this` in functions passed to `test()` and `beforeEach()` and whatnot. If you prefer to scope any of this state to "the currently running test" you could store it on that object instead of module-scoping it. I guess you'd want fallbacks in case the user is running against the test API, and I'm not sure what you want to do if a test isn't currently running...anyway, not certain this is a direction you want to go, but figured I'd mention it in case.

bendemboski: Oh, I guess this code only runs when `Ember.testing` is true, so you at least don't have to worry about the using-new-api-but-test-not-running case.

addon/testing.js
```diff
@@ -0,0 +1,73 @@
...
+        if (this.hasResumed) {
+          return;
+        }
+        this.hasResumed = true;
+        taskInstance.proceed(resumeIndex, YIELDABLE_CONTINUE, null);
+      },
+    });
+
+    return () => {
+      if (intervalId) {
```
bendemboski: I must be missing something here...it doesn't look like `intervalId` ever gets set...

machty: yeah that was something i started in on and forgot to remove

addon/testing.js
machty: I def didn't know about getContext(), that's a much better place for it :)

gabrielgrant:
Is the currently-broken tagged timers test an expected failure? (wondering what parts of this, if any, i should be trying out?)

machty:
@gabrielgrant yes i'm sorry it's totally broke right now, will ping you when it's more testable (that said we've had some convos in `#e-concurrency` slack channel that make it seem more likely that this PR will be closed in favor of another lib like ember-milestones)

gabrielgrant:
Cool, hadn't heard of ember-milestones. I'll join `#e-concurrency` to follow the discussion there

machty:
Closing this in favor of recommending [ember-milestones](https://salsify.github.io/ember-milestones/latest/docs/) as a superior approach

PR #218, opened 2018-04-13
`toFunction` modifier

maxfierke:
Cherry-picked the `toFunction` bits from @machty's `to-function` branch.

`toFunction`, does just that, takes a task and gives you back the performable task as a function. Super cool and makes it possible to attach tasks to POJOs/non-EmberObjects.

#### Open questions:
* Do we still need to use the `WeakMap` polyfill?
* Would it make sense to allow `toFunction` to take a single parameter that could be used for the concurrency context? i.e. rather than using the same default for functions not associated with an object, allow the user to decide what context to use.

#### Todo
* [x] Docs


buschtoens:
Related to #240. Didn't see this before.

maxfierke:
Closing, as there will probably be a better solution coming out of #302 and/or #289

PR #223, opened 2018-04-20
Add TaskInstance events API

maxfierke:
Adds a number of task instance lifecycle events that are fired
on the host object:

- started: fired when a new task instance has started executing
- succeeded - fired when a task instance completes successfully
- errored - fired when a task instance throws an uncaught error
- canceled - fired when a task instance is canceled

e.g.

```js
import { Component } from '@ember/component';
import { on } from '@ember/object/evented';
import { inject as service } from '@ember/service';

export default Component.extend({
  notify: service(),
  doSomething: task(function * () {
    // ... does something that might throw
  }),

  handleTaskError: on('doSomething:errored', function (taskInstance, error) {
    this.get('notify').error(`Captain, we have problem: ${error.message}`);
  })
});
```

Further documentation on use is provided in the JSDoc annotations and in the documentation site.

### DIFF
--- a/addon/-property-modifiers-mixin.js
+++ b/addon/-property-modifiers-mixin.js
@@ -14,6 +14,7 @@ export const propertyModifiers = {
   _taskGroupPath: null,
   _hasUsedModifier: false,
   _hasSetBufferPolicy: false,
+  _hasEnabledEvents: false,
 
   restartable() {
     return setBufferPolicy(this, cancelOngoingTasksPolicy);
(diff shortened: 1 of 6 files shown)


machty:
This looks great!

I am a bit worried though that adding this functionality would introduce a performance hit since every task is going to start firing events on the parent object even when 99% of the time no one's going to be listening for those events. How would you feel instead about requiring tasks be "annotated" with `.evented()` to enable them firing events on the host object?

e.g.

```
export default Component.extend({
  notify: service(),
  doSomething: task(function * () {
    // ... does something that might throw
  }).evented(),

  handleTaskError: on('doSomething:errored', function (taskInstance, error) {
    this.get('notify').error(`Captain, we have problem: ${error.message}`);
  })
});
```

And in the future we can consider making `.evented()` the default if task events become popular and it can be shown that there's not a huge perf hit?

maxfierke:
@machty sure thing. I can dig it.

PR #224, opened 2018-04-25
Allow labels on timeouts for debugging

XuluWarrior:
The RSVP Promise API has an additional parameter for setting the label on the created Promise.
The ember-inspector extension displays these labels in the Promises view which is useful for working out what Promises are still pending.


machty:
I'm hesitant to merge this because there have been other proposal for what the additional arguments to `timeout()` should be (such as a proposal for tagging the timeouts so that they can be skipped/paused while testing).

XuluWarrior:
Although being able to label timeouts would be useful to us, it's not a major issue and your other proposals sound good as well.  I will wait to see which way you choose to go.

PR #248, opened 2018-09-21
Fix ES5 getter syntax on Ember Canary

machty:
Root cause of issue:

https://github.com/emberjs/ember.js/commit/246435e7260f45d9e5cc35a4a8b4de4f011f3625

The above commit started making use of the previously
unused Descriptor#setup method that e-c uses
for setting up observers and e-c wasn't calling
the super method, which was causing the ES5
getter to not be set on the object.

This commit calls the super `setup` method if it
exists so that Object.defineProperty is properly
called.

/cc @bendemboski

### DIFF
--- a/addon/-task-property.js
+++ b/addon/-task-property.js
@@ -435,11 +435,16 @@ export function TaskProperty(taskFn) {
   this._observes = null;
 }
 
+const superSetup = _ComputedProperty.prototype.setup;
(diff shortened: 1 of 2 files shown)


bendemboski:
@machty confirmed, it fixed my repo (sorry, Discord connection issues are keeping me from replying over there)

Issue #243, opened 2018-09-01
RFC: Reset derived state


amiel:
## Summary

In this issue, I would like to propose public API to reset the derived state
for a task. Calling the function would cause the `last*` task object properties
to reset to `undefined`, allowing interface that depends on these properties to
"start over".

## Motivation

The motivation for this change is best described with an example. I know I've
looked for the existence of this functionality in the documentation for other
use-cases. I have chosen this use case as my example as it is what I am working
on right now and is therefore freshest.

I have a form, where at first there is one dropdown element, and when an item
from the first dropdown is selected, an ajax request is fired to load options
for a second element. In my current case, there is also a third dropdown, whose
options get loaded when an item is selected from the second dropdown.

The following will use country, region, and city, as an arbitrary and hopefully
understandable example. Below are a component template and implementation to
illustrate how I accomplish this now. Derived state is reset by writing the
task to return `undefined`, thereby unsetting the value from `lastSuccessful`.

<details>
  <summary>Template</summary>

  ```hbs
  {{ember-power-select options=countries onChange=(action 'selectCountry')}}

  {{! Not shown for simplicity: a loading indicator when a task isRunning }}

  {{#if regions}}
    {{ember-power-select options=regions onChange=(action 'selectRegion')}}
  {{/if}}

  {{#if cities}}
    {{ember-power-select options=cities onChange=(action (mut 'city'))}}
  {{/if}}
  ```

</details>

<details>
  <summary>Component</summary>

  ```js
  import Component from '@ember/component';
  import { readOnly } from '@ember/object/computed';
  import { task } from 'ember-concurrency';

  export default Component.extend({
    countries: [...],

    regions: readOnly('loadRegions.lastSuccessful.value'),
    cities: readOnly('loadCities.lastSuccessful.value'),

    loadRegions: task(function*(country) {
      return yield ajax(...);
    }),

    loadCities: task(function*(region) {
      // When called with an explicitly falsy value, return undefined to reset
      // derived state.
      if (!region) { return; }

      return yield ajax(...);
    }),

    actions: {
      selectCountry(country) {
        this.set('country', country);

        // If a region and city have been selected, and the country is changed,
        // we need to reset the state of the cities, or the cities dropdown
        // will still be populated with options from the previously selected
        // country. This explicitly resets the derived state for cities.
        this.get('loadCities').cancelAll();
        this.get('loadCities').perform(null);

        this.get('loadRegions').perform(country);
      },

      selectRegion(region) {
[...]

PR #253, opened 2018-10-18
cancelAll({ resetState: true }) can remove existing derived state

amiel:
This is a first attempt at https://github.com/machty/ember-concurrency/issues/243.

This change includes two changes:

1. Change the signature of `cancelAll(reason)` to `cancelAll({ reason })`

   It was stated in https://github.com/machty/ember-concurrency/issues/243#issuecomment-424844648 that it would probably be ok to change this without backwards compatibility. However, if there's a desire to have backwards compatibility, it wouldn't be hard to add.

   Also note that I changed this signature for both the `cancelAll` on the scheduler as well as the task property. It isn't actually necessary to change the signature on the scheduler, but it seemed worth being consistent there. Again, let me know if you have a strong opinion :man_shrugging: 

2. Add a `resetState` option to `taskProperty.cancelAll`

   Currently, this explicitly sets state attributes to `null`, effectively throwing away any existing state. I have almost no understanding of the internals of ember-concurrency, so I have no idea what consequences this might have. Please review and let me know if there might be a better approach to take here.

### DIFF
--- a/addon/-scheduler.js
+++ b/addon/-scheduler.js
@@ -23,7 +23,7 @@ const Scheduler = EmberObject.extend({
     this.queuedTaskInstances = [];
   },
 
-  cancelAll(reason) {
+  cancelAll({ reason }) {
     let seen = [];
     this.spliceTaskInstances(reason, this.activeTaskInstances, 0, this.activeTaskInstances.length, seen);
     this.spliceTaskInstances(reason, this.queuedTaskInstances, 0, this.queuedTaskInstances.length, seen);
--- a/addon/-scheduler.js
+++ b/addon/-scheduler.js
@@ -23,7 +23,7 @@ const Scheduler = EmberObject.extend({
     this.queuedTaskInstances = [];
   },
 
-  cancelAll({ reason }) {
+  cancelAll(reason) {
     let seen = [];
     this.spliceTaskInstances(reason, this.activeTaskInstances, 0, this.activeTaskInstances.length, seen);
     this.spliceTaskInstances(reason, this.queuedTaskInstances, 0, this.queuedTaskInstances.length, seen);
--- a/addon/-task-state-mixin.js
+++ b/addon/-task-state-mixin.js
@@ -47,14 +47,7 @@ export default Mixin.create({
     this._scheduler.cancelAll({ reason });
 
     if (resetState) {
-      this.setProperties({
-        'last': null,
-        'lastRunning': null,
-        'lastPerformed': null,
-        'lastSuccessful': null,
-        'lastComplete': null,
-        'performCount': 0,
-      });
+      this._resetState();
     }
   },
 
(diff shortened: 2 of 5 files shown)


addon/-task-state-mixin.js
```diff
@@ -50,5 +57,18 @@ export default Mixin.create({
...
+  _resetState() {
+    this.setProperties({
+      'last': null,
+      'lastRunning': null,
+      'lastPerformed': null,
+      'lastSuccessful': null,
+      'lastComplete': null,
+      'lastErrored': null,
+      'lastCanceled': null,
+      'lastIncomplete': null,
```
amiel: Would it make any sense to not reset `lastErrored`, `lastCanceled`, and/or `lastIncomplete`?

addon/-scheduler.js
```diff
@@ -23,7 +23,7 @@ const Scheduler = EmberObject.extend({
     this.queuedTaskInstances = [];
   },
 
-  cancelAll(reason) {
+  cancelAll({ reason }) {
```
machty: Any reason (lol) this needed to be changed?

amiel: Not really. As mentioned in my comment, I did this for internal consistency (and when I started with the signature change, I honestly wasn't sure which implementation of `cancelAll` would get the `resetState` logic).

I would be happy to revert the signature change of `scheduler.cancelAll` if you like. What do you think?

machty: Yeah it's not a big thing, but since `cancelAll` on the scheduler is pretty private, I'd prefer a smaller diff.

addon/-task-state-mixin.js
machty: I think it's best to reset everything back to the state when the task was first initialized; it'd also be easy to forget which properties _didn't_ get reset and why.

amiel: That makes sense. Just to verify, you are suggesting I leave this as it is now?

addon/-scheduler.js
amiel: I will make the change :)

amiel:
I think I've addressed your review so this should be ready for re-review. Thanks!

machty:
This looks great, I believe I'll merge this in a day or two pending any further last minute feedback from interested parties.

addon/-task-state-mixin.js
```diff
@@ -50,5 +57,18 @@ export default Mixin.create({
 
   _scheduler: null,
 
+  _resetState() {
+    this.setProperties({
```
bendemboski: Shouldn't this reset `lastStarted` too?

machty: Oops, missed this this... yes it probably should. I'll make this adjustment on a separate commit and release in the next version.

machty:
Released in 0.8.24

amiel:
Thanks @machty!

machty:
Thanks for working on it!

erichonkanen:
is this in the docs?

amiel:
@erichonkanen It's not in the docs, but it should be. Do you think a casual mention on the cancelation page would be enough?

erichonkanen:
just ran into this again on a separate project 😆 yeah that'd be nice to add to the docs, although now I remembered to check if clearing the state was possible yet!

PR #254, opened 2018-10-18
Add integrity hashes in yarn.lock

amiel:
While working on https://github.com/machty/ember-concurrency/pull/253, I followed the instructions to set up development (namely run `yarn install`). When I ran `yarn install`, yarn added these hashes for integrity. It looks like [this feature](https://github.com/yarnpkg/yarn/pull/5042) was added to yarn in [1.10.0](https://github.com/yarnpkg/yarn/releases/tag/v1.10.0). Feel free to take this PR or leave it. If you don't want to add this change, it might be worth changing the README to include an option to yarn to prevent this.


jherdman:
@amiel FYI these values appear to be in the yarn.lock file. I think you can safely close this now.

PR #266, opened 2018-12-27
fix: #237 - replacing `Ember.Logger` with `console`

### LINKED ISSUE #237: Ember.Logger is deprecated
This addon makes use of Ember.Logger, which [is now deprecated](https://emberjs.com/deprecations/v3.x#toc_use-console-rather-than-ember-logger).

cs3b:
closes #237


machty:
Thank you!

PR #270, opened 2019-01-13
correct error message for later versions of Ember

BryanCrotaz:
When you create a task manually using `set(obj, 'myTask', task(function*(){})` and then `perform` it the error message is misleading. This is the same issue as the deprecation added recently to computed properties, so I've used the same deprecation warning.


BryanCrotaz:
Something outside the scope of my PR is failing:

```
error ip-regex@3.0.0: The engine "node" is incompatible with this module. Expected version ">=8". Got "6.16.0"
error Found incompatible module
```

Issue #244, opened 2018-09-03
'run to completion' error if exception thrown in task


BryanCrotaz:
@restartableTask
*_debounce(this:MyClass) {
  yield timeout(250);
  this._doWork();
}
This gives an assertion error if `this._doWork` throws. In my case it was using a non existent variable.
`Assertion Failed: The task generator function has already run to completion. This is probably an ember-concurrency bug.`

machty:
@BryanCrotaz can you reproduce in a [Twiddle](https://ember-twiddle.com/b2b0c016f4df24261381487b60c707f3)?

BryanCrotaz:
No. It gives the correct missing variable error.
My guess is it's some combination of typescript/ES6/ember-concurrency-decorator.

BryanCrotaz:
If I fix the missing variable typo in my code the problem goes away.

buschtoens:
Are you using ember-cli-typescript or ember-cli-babel with TypeScript compatibility?

Is it _only_ with `@restartableTask` or `@task` as well?

What about...?

```ts
@restartableTask
_debounce = function*(this: MyClass) {
  yield timeout(250);
  this._doWork();
}
```

I know that Babel had / has trouble with decorating generator methods, as in your example. The initializer form, as shown above, has always worked so far. 

If this is not the cause, maybe it's the this type?

PR #276, opened 2019-02-06
added tests for throwing during a task

BryanCrotaz:
Trying (and so far failing) to reproduce #244

### DIFF
--- a/tests/unit/error-handling-test.js
+++ b/tests/unit/error-handling-test.js
@@ -169,26 +169,34 @@ module('Unit: task error handling', function() {
     assert.equal(state, 2, "continued after error");
   });
 
-  test("task that catches throw in yield block should continue", async function(assert) {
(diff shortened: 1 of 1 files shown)


BryanCrotaz:
Found it.

The sequence that fails:

try
    yield something that throws
catch (e) 
    ignore the error, carry on

destroy the object

yield something else

machty:
image

Running the test locally and it seems the only error is the `try { } catch() {}` in the async test function, which is raising an error because async functions don't know how to "handle/await" promises that resolve to TaskCancelations other than to treat them like errors. While [e-c's handling of cancelation can cause confusion](http://ember-concurrency.com/docs/error-vs-cancelation), I don't believe it is the source of your issue, and I don't think your breaking test demonstrates the same `The task generator function has already run to completion` error you're seeing.

Issue #216, opened 2018-04-10
`return yield taskInstanceThatThrows` behaves differently than `return taskInstanceThatThrows`


machty:
Noticed this when trying out e-c-retryable

```
  fetchAllDependencies: retryable(task(function*() {
    return all([
      this.get('loadAdyen').unlinked().perform(),
      this.get('initiatePayment').unlinked().perform(),
      this.get('liquidFireTransitions').waitUntilIdle(),
      this.get('testThrow').perform(),
    ]);
  }), retryPolicy),
```

This ^ somehow escapes `retryable`'s control and throws Ember.onerror; if I add `return yield all` it works. In this case, the semantics should be the same.

maxfierke:
Tried to reproduce this with a couple tests to see if this was an e-c-retryable bug or an e-c bug: https://gist.github.com/maxfierke/107474a159b8d0b10cb93b860cd1233e

Not sure if these test cases match exactly what's happening with these tasks, but it seems that with plain e-c, both `all` and `yield` all will get caught by `onerror`, while with e-c-retryable works as you say with `yield all`. Assuming the test cases are valid for this bug, then e-c-retryable must be fixing it accidentally for the yield case somehow.

machty:
@maxfierke would you mind setting up a Twiddle that demonstrates what you're seeing?

maxfierke:
@machty ported those test cases over to a Twiddle: https://ember-twiddle.com/f714948db37fac99c8ad02ea5d520f67?numColumns=2&openFiles=controllers.application.js%2Ccontrollers.application.js

As you can see with the last one, the error does not get bubbled to `Ember.onerror`

maxfierke:
It might be due to the way e-c-retryable is slightly altering when the task runs due to the way that it wraps & executes the task function here: https://github.com/maxfierke/ember-concurrency-retryable/blob/ca640543e7236ee312e99f4463232a95e412f6ce/addon/retryable.js#L16, using `return yield*`

But this is interesting... if the task that throws has a yield before it throws, they all seem to behave the same: https://ember-twiddle.com/e62be2d9ce7d72a92480d27b4ebb3865?openFiles=controllers.application.js%2C

So, if an error is thrown in the first chunk of a task before the first `yield`, then it seems it hits `Ember.onerror`, before the `catch` surrounding it. Whereas with the timeout, the `catch` gets hit but `Ember.onerror` does not (unless you click it really fast :p, but that's something different)

[...]

PR #277, opened 2019-02-09
[BUGFIX] Updates internals for Ember Canary

pzuraq:
We just landed some pretty major changes to the Descriptor system
upstream. We verified that ember-concurrency would still be able to work
first, this branch is that work. We should likely update it to use
`ember-compatibility-helpers` as well, so you can continue to support
older versions of Ember at the same time.

### DIFF
--- a/addon/-task-group.js
+++ b/addon/-task-group.js
@@ -2,11 +2,8 @@ import { or, bool } from '@ember/object/computed';
 import EmberObject from '@ember/object';
 import { objectAssign, _ComputedProperty } from './utils';
 import TaskStateMixin from './-task-state-mixin';
-import {
-  propertyModifiers,
-  resolveScheduler
-} from './-property-modifiers-mixin';
-
+import { propertyModifiers } from './-property-modifiers-mixin';
+import { gte } from 'ember-compatibility-helpers';
 
 export const TaskGroup = EmberObject.extend(TaskStateMixin, {
   isTaskGroup: true,
@@ -17,24 +14,15 @@ export const TaskGroup = EmberObject.extend(TaskStateMixin, {
 
   _numRunningOrNumQueued: or('numRunning', 'numQueued'),
   isRunning: bool('_numRunningOrNumQueued'),
-  isQueued:  false
+  isQueued: false,
 });
 
-export class TaskGroupProperty extends _ComputedProperty {
-  constructor(taskFn) {
-    let tp;
-    super(function(_propertyName) {
-      return TaskGroup.create({
-        fn: taskFn,
-        context: this,
-        _origin: this,
-        _taskGroupPath: tp._taskGroupPath,
-        _scheduler: resolveScheduler(tp, this, TaskGroup),
-        _propertyName,
-      });
-    });
-    tp = this;
-  }
+export let TaskGroupProperty;
+
+if (gte('3.10.0')) {
+  TaskGroupProperty = class {};
+} else {
+  TaskGroupProperty = class extends _ComputedProperty {};
 }
 
 objectAssign(TaskGroupProperty.prototype, propertyModifiers);
--- a/addon/-task-group.js
+++ b/addon/-task-group.js
@@ -19,7 +19,7 @@ export const TaskGroup = EmberObject.extend(TaskStateMixin, {
 
 export let TaskGroupProperty;
 
-if (gte('3.10.0')) {
+if (gte('3.10.0-alpha.1')) {
   TaskGroupProperty = class {};
 } else {
   TaskGroupProperty = class extends _ComputedProperty {};
--- a/addon/-task-property.js
+++ b/addon/-task-property.js
@@ -3,29 +3,23 @@ import { addObserver } from '@ember/object/observers';
 import { addListener } from '@ember/object/events';
 import EmberObject from '@ember/object';
 import { getOwner } from '@ember/application';
-import {
-  default as TaskInstance,
-  getRunningInstance
-} from './-task-instance';
+import { default as TaskInstance, getRunningInstance } from './-task-instance';
 import {
   PERFORM_TYPE_DEFAULT,
   PERFORM_TYPE_UNLINKED,
-  PERFORM_TYPE_LINKED
+  PERFORM_TYPE_LINKED,
 } from './-task-instance';
 import TaskStateMixin from './-task-state-mixin';
-import { TaskGroup } from './-task-group';
-import {
-  propertyModifiers,
-  resolveScheduler
-} from './-property-modifiers-mixin';
+import { propertyModifiers } from './-property-modifiers-mixin';
 import {
   objectAssign,
   INVOKE,
   _cleanupOnDestroy,
-  _ComputedProperty
+  _ComputedProperty,
 } from './utils';
 import EncapsulatedTask from './-encapsulated-task';
 import { deprecate } from '@ember/debug';
+import { gte } from 'ember-compatibility-helpers';
 
 const PerformProxy = EmberObject.extend({
   _task: null,
@@ -33,7 +27,11 @@ const PerformProxy = EmberObject.extend({
   _linkedObject: null,
 
   perform(...args) {
-    return this._task._performShared(args, this._performType, this._linkedObject);
+    return this._task._performShared(
+      args,
+      this._performType,
+      this._linkedObject
+    );
   },
 });
 
@@ -184,10 +182,15 @@ export const Task = EmberObject.extend(TaskStateMixin, {
     if (typeof this.fn === 'object') {
       let owner = getOwner(this.context);
       let ownerInjection = owner ? owner.ownerInjection() : {};
-      this._taskInstanceFactory = EncapsulatedTask.extend(ownerInjection, this.fn);
+      this._taskInstanceFactory = EncapsulatedTask.extend(
+        ownerInjection,
+        this.fn
+      );
     }
 
-    _cleanupOnDestroy(this.context, this, 'cancelAll', { reason: 'the object it lives on was destroyed or unrendered' });
+    _cleanupOnDestroy(this.context, this, 'cancelAll', {
+      reason: 'the object it lives on was destroyed or unrendered',
+    });
   },
 
   _curry(...args) {
@@ -315,7 +318,6 @@ export const Task = EmberObject.extend(TaskStateMixin, {
    * @readOnly
    */
 
-
   /**
    * The current number of active running task instances. This
    * number will never exceed maxConcurrency.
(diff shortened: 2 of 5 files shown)


bendemboski:
Hey @pzuraq can you give me some advice/help on getting `ember-concurrency-test-waiter` to work with this update? Currently it kinda just monkey-patches task properties [here](https://github.com/bendemboski/ember-concurrency-test-waiter/blob/8d302fbb24899c8ee59a05f5da485f0254e3f91f/addon/with-test-waiter.js#L12-L29), but that will no longer work.

So what I'm trying to figure out is a solution that will work with `ember-concurrency` both before and after this change, possibly taking this PR as an opportunity to provide a cleaner hook for what `ember-concurrency-test-waiter` is trying to do.

pzuraq:
Ah, that's interesting. I think the only way for us to support this behavior would actually be to continue to add the `taskFn` to the decorator as a property, then have you replace it like you are currently 😕 we could also add it to a `WeakMap` potentially and make that map accessible, I'm not sure if that's much cleaner though

bendemboski:
Well, it's kinda ugly and I'm not happy about how it encodes an assumption about the arguments that the framework passes to `tp`, but I found a way to make `ember-concurrency-test-waiter` work with this PR -- https://github.com/bendemboski/ember-concurrency-test-waiter/commit/fc119c529574554278063b8348c0b9ac30e37a46.

@pzuraq I'm curious to hear your thoughts -- is there a cleaner way to effectively wrap a computed property/descriptor?

I suppose another approach would be to instrument the task property. Maybe:

* insert `let taskFnWrappers = [];` [here](https://github.com/machty/ember-concurrency/pull/277/files#diff-7562eb8eba3c3d86a671045f8c70107eR59)
* insert `tp.addTaskFnWrapper = function(fn) { taskFnWrappers.push(fn); }` [here](https://github.com/machty/ember-concurrency/pull/277/files#diff-7562eb8eba3c3d86a671045f8c70107eR125) that allows modifiers to register wrapper functions that are called with the task function (bound to its arguments) so they can run code before/after the task function.
* add some code [here](https://github.com/machty/ember-concurrency/pull/277/files#diff-7562eb8eba3c3d86a671045f8c70107eR62) to wrap nest/wrap all the wrapper functions around the task function and pass that wrapped function to the task instance.

What are your thoughts on which approach would be best?

cc: @machty

bendemboski:
Okay, I hashed out my second proposal, and I kinda like how it turned out. [Here](https://github.com/bendemboski/ember-concurrency/commit/fce47a814ec069acdb9050e7c1c18997b0067432) is the proposed change to this PR, and [here](https://github.com/bendemboski/ember-concurrency-test-waiter/commit/ff775dd32a0c7bf0f79b0c825ed607e76616f6de) is the corresponding change to `ember-concurrency-test-waiter` (and this pattern could theoretically be used by other addons that want to wrap `ember-concurrency` task functions if there ever are any).

I think it's nice and clean and doesn't violate encapsulation or encode brittle assumptions about how Ember (or even `ember-concurrency`) internals work. I imagine there are other abstractions that could accomplish the same thing, and I'm certainly open to them, but if folks are happy with this, maybe we can move forward with it?

@pzuraq I have some time to work on moving this whole thing forward if you can help me with:

1. Figuring out how to use `ember-compatibility-helpers` (what feature do we check for? does it exist yet, or does `ember-compatibility-helpers` need a PR too?)
2. Figuring out how to structure the code to be cross-compatible -- do I just take the whole diff in this commit and wrap every substantive change in an if block checking `ember-compatibility-helpers` (and then maybe organizing the code a little better, like putting the old code in a single file of helpers or something)? Or can you envision a cleaner way to organize the two mechanisms?
3. General support/help with questions, as there's a lot here I haven't learned about yet, so stuff might come up!

pzuraq:
I'm ok with that proposal, I would like to see what @machty thinks since it would essentially be a new public API and may incur some runtime cost. The nice thing about using a `WeakMap` to get/set the function is it would be possible to strip that out in production builds, and keep it a private/intimate API.

For your remaining questions:

1. We've mostly been using the `gte`/`lte` flags from compatibility-helpers, they allow you to specify a version like so:

```js
if (gte('3.9.0')) {

} else {

}
```

  The first code block will run if the version is 3.9.0+, including betas and canaries. The additional flags were harder to maintain since there are so many of them, so we aren't really using them anymore.

2. You can take a look at [ember-decorators](https://github.com/ember-decorators/ember-decorators/blob/master/packages/utils/addon/computed.js#L22) to see how we handled it there, but generally yes. Sometimes it's possible to just tweak one or two things depending on different versions, but in this case we found it really required entirely different implementations.

3. Absolutely! Feel free to ping me on Discord about this, or ask more questions in this issue 😄 I'll also give you commit to my fork so you can iterate there if you'd like

machty:
I will try to catch up on this over the weekend.

Could someone please summarize the main thing(s) that changed about Descriptors that are impacting ember-concurrency?

pzuraq:
Descriptors are now spec compliant decorator functions, and have changed significantly under the hood. We have a new implementation that works, but it will be incompatible with older versions of Ember. This won’t be a problem as we can use ember-compatibility-helpers to ship the correct version.

However, a sideeffect of this is making most of the implementation details of descriptors much more private, and in turn making more of the implementation details of ember-concurrency more private. @bendemboski’s addon was relying on these private implementation details to create a test-waiter, so we need to figure out how to make them public-ish, or make it clear it’s not a supported use case.

machty:
Still haven't dug in deeply but it seems like it's making more and more sense to have an official ember-concurrency middleware for decorating/wrapping/extending task functions, so that separate e-c plugins don't have to know anything about `ember-compatibility-helpers`. 

Random thought but I believe there's an issue with trying to wrap a task generator fn with `yield * original`; there's some cases where I think if your wrapped task fn returns a promise/task instance that throws an error, the error isn't properly handled by e-c. I noticed this ember-concurrency-retryable cc @maxfierke . This is another reason to have a sanctioned e-c wrapping/extending API.

maxfierke:
yes, it subtly alters how the task runs. Basically, in vanilla e-c right now (no wrapping), the part before the first `yield` will run immediately, and errors will not be caught/handled by e-c. Whereas, wrapping will run it within the e-c scheduler, and be handled in e-c (I think maybe it runs w/ the next microtasks?). I think this was essentially how it worked in ember-concurrency before 0.8.x, IIRC. (Examples are here: https://github.com/machty/ember-concurrency/issues/216#issuecomment-390350154, for anyone interested)

It would be a subtly breaking change to wrap by default, but it's one that's already being incurred via either `ember-concurrency-retryable` or possibly via `ember-concurrency-test-waiter`. It seems like an Ben's approach to an API for taskFn wrapping would be a pragmatic way of solving this (until there's a more concrete idea how something like e-c middleware would work) provided that it did no wrapping if not invoked.

pzuraq:
Status update here: Currently waiting on some pending changes to the implementation, it looks like we may revert it from the 3.9 beta and wait until 3.10 to land it due to that, give us some more time to handle things. Should know by the core team meeting on Friday!

rwjblue:
FYI - we (but really @pzuraq) did revert the changes for 3.9 (current beta) over in https://github.com/emberjs/ember.js/pull/17649, to buy us more time to experiment with the feature on canary. Thank you @pzuraq!

bendemboski:
@pzuraq @rwjblue any idea when the reversion will land in the `ember-beta` channel that `ember-try` uses? I'm having to disable the `ember-beta` scenario in more and more addons that use `ember-concurrency`...

pzuraq:
Just checked, it's looking like it'll be cut later tonight, but definitely by tomorrow at the latest. Sorry for the delays!

pzuraq:
Beta.2 is out, should be fine!

pzuraq:
This is ready to review!

addon/-task-group.js
```diff
@@ -17,24 +14,15 @@ export const TaskGroup = EmberObject.extend(TaskStateMixin, {
...
-        _taskGroupPath: tp._taskGroupPath,
-        _scheduler: resolveScheduler(tp, this, TaskGroup),
-        _propertyName,
-      });
-    });
-    tp = this;
-  }
+export let TaskGroupProperty;
+
+if (gte('3.10.0')) {
```
gossi: Mustn't it be `gte('3.10.0-beta.1')` I remember robert once told me ?

gossi:
I just tried your PR with the latest canary:

```
DEBUG: -------------------------------
DEBUG: Ember      : 3.10.0-canary+40ecde67
DEBUG: Ember Data : 3.8.0
DEBUG: -------------------------------
```

I had errors before, that went away -> #worksforme ;)

addon/-task-group.js
pzuraq: We actually strip off everything past the main semver version in `ember-compatibility-helpers` for canary, since canary appends the actual commit sha and completely breaka these functions.

rwjblue: Right, but we don’t strip the beta bits (and if we are we shouldn’t 😜). I would do:

```suggestion
if (gte('3.10.0-alpha.1’)) {
```

pzuraq: We do, but we should deprecate it upstream. I definitely think some users rely on that behavior, but agree it’s not ideal.

buschtoens:
I can also confirm that https://github.com/machty/ember-concurrency-decorators/pull/39 works with this branch! 🎉 
https://travis-ci.org/machty/ember-concurrency-decorators/builds/502538049

machty:
I still don't deeply understand the new vs old mechanics and limitations of descriptors and how that's impacted e-c, but this PR looks good and I'm happy to merge.

@maxfierke does this PR still risk introducing the execution corner case bugs mentioned earlier?

@bendemboski does this PR now work with your e-c-test-waiters addon?

bendemboski:
Yeah, thanks to [this](https://github.com/machty/ember-concurrency/pull/277/files#diff-7562eb8eba3c3d86a671045f8c70107eR90) and related changes (and @pzuraq!) this branch works with `ember-concurrency-test-waiters` on canary and pre-canary https://travis-ci.org/bendemboski/ember-concurrency-test-waiter 🎉

maxfierke:
@machty nope, should be okay, as `taskFn` is passed right on through with this PR. My comment about about the execution corner cases applied only to one of the proposals in here for a `taskFn` wrapping API, if `taskFn` were to no-longer be available on the `TaskProperty` directly.

If it works for `ember-concurrency-test-waiters`, it should also work for `ember-concurrency-retryable` too. I'll test with it tomorrow to be sure, once I have access to reliable enough internet to clone the branch.

PR #282, opened 2019-03-15
Use correct task name in snippets

danwenzel:
The task name is `queryServer`, but `fetchResults` was referencing `doStuff`


machty:
Thank you!

Issue #240, opened 2018-07-24
Getting rid of `.perform()` in JS


buschtoens:
#### tl;dr

```js
// I want
this.someTask('🥦');

// instead of
this.someTask.perform('🥦');
```

---

#### We already took the first step

This nifty trick allows users to just use the `{{action}}` helper with tasks in their templates:

https://github.com/machty/ember-concurrency/blob/5c830b1dfdc9609b4de841b503bdb36023bbb4b1/addon/-task-property.js#L395-L397

You'd use it as:

```hbs
<button onclick={{action deleteUser user}}>
  Delete {{user.name}}
</button>
```

This is super sexy. 🔥

The templates are agnostic to whether or not the thing that you're passing is a regular (closure) action or a task.

I would _love_ to have the equivalent in JS and get rid of `.perform()`.

#### It's already improved in JS

Since ES5 getters landed, ergonomics already improved massively for tasks, because instead of this:

```js
export default Ember.Controller.extend({
  someTask: task(function*() {
    console.log('😐');
  }),
  
  actions: {
    performSomeTask() {
      this.get('someTask').perform();
    }
  }
});
```

You can know just go ([Ember Twiddle](https://ember-twiddle.com/035cbe7fe7e680cf0cee6b875d3e02c8?openFiles=controllers.application.js%2C)):

```js
export default Ember.Controller.extend({
  someTask: task(function*() {
    console.log('😍');
  }),
  
  actions: {
    performSomeTask() {
      this.someTask.perform();
    }
  }
});
```

#### One last step

But what's still annoying is that, when you refactor a simple method to a task, you have to update all the places that call this method, because you need to switch from:

```js
this.someMethod('🥦');
```

to:

```js
this.someMethod.perform('🥦');
```

#### What would it take?

The [`TaskProperty`](http://ember-concurrency.com/api/TaskProperty.html), when it is `.get()`ed should not return an object that is a [`Task`](http://ember-concurrency.com/api/Task.html), but instead should return a function that has all the `Task` stuff on its prototype. There might be some mad science involved, but assuming that

a) this does not hurt performance
b) we do not remove `.perform()` (for now) to avoid API churn
c) it works reliably

would you be open to this change? Then I might try my hand at a PR and see if this even works.

maxfierke:
I know there was a discussions about this on #e-concurrency, but totally forgot what was discussed, and now it's lost to Slack history ¯\\_(ツ)_/¯

[...]

PR #289, opened 2019-04-09
feat: plain task invocation

buschtoens:
Implements #240.

For some reason changing `numRunning` and `numQueued`, if accessed via `taskInstance.task`, do not trigger an update. If accessed via `taskInstance.owner[taskInstance.task.name]`, it works. This has to be a problem with the tracking system.

https://github.com/machty/ember-concurrency/blob/62c942d7f41e1555ab835ba921f1293f0c8d1984/addon/-scheduler.js#L94-L97

Unfortunately `Acceptance | injections on encapsulated tests: encapsulated tasks support injections` is also failing.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -19,7 +19,7 @@ function _computed(fn) {
       return computed(fn)(...arguments);
     };
 
-    Ember._setComputedDecorator(cp);
+    Ember._setClassicDecorator(cp);
 
     return cp;
   } else {
@@ -75,7 +75,7 @@ function _computed(fn) {
 export function task(taskFn) {
   let tp = _computed(function(_propertyName) {
     tp.taskFn.displayName = `${_propertyName} (task)`;
-    return Task.create({
+    const task = Task.create({
       fn: tp.taskFn,
       context: this,
       _origin: this,
@@ -83,8 +83,17 @@ export function task(taskFn) {
       _scheduler: resolveScheduler(tp, this, TaskGroup),
       _propertyName,
       _debug: tp._debug,
-      _hasEnabledEvents: tp._hasEnabledEvents,
+      _hasEnabledEvents: tp._hasEnabledEvents
     });
+
+    if (!gte('3.10.0')) {
+      return task;
+    }
+
+    const perform = (...args) => task.perform(...args);
+    Object.setPrototypeOf(perform, task);
+
+    return perform;
   });
 
   tp.taskFn = taskFn;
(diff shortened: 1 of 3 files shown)


buschtoens:
It seems that Glimmer is not happy about `Object.setPrototypeOf` in general. I've made a repro here: https://codesandbox.io/s/6z8n0qyj0w?initialpath=%2Ftests

NullVoxPopuli:
where is this at? anyway I can help?

maxfierke:
Closing this out. As awesome as this experiment is/was, this PR targets v0/v1 of ember-concurrency, so it would need to be re-worked before it could be merged. `Object.setPrototypeOf` is still a slow path in engines so we'd also probably need to do something at the base `Task` class level to avoid problems.

dfreeman:
> Object.setPrototypeOf is still a slow path in engines

@maxfierke I'd love to hear more about this. I know it can absolutely tank your performance when you change the prototype of objects that are already 'out in the world', but for fresh objects it's less clear (to me, anyway) what the implications are.

Microbenchmarks are meaningless on their own, and it's also possible I've messed something up, but I put together [a benchmark for a few implementations for a tiny task-like entity](https://jsbench.me/6oktwyhdbw/1): one with `perform()`, one callable with `defineProperties` for derived state, and one callable with `setPrototypeOf` for derived state.

In SpiderMonkey (Firefox) and JavaScriptCore (Safari), the difference in overall performance for the three is relatively small—`perform()` does a few percent better on average in Firefox and a few percent worse in Safari. In V8, though, the `setPrototypeOf` implementation surprisingly did __40% better__ than the other two.

Of course this doesn't mean much in isolation, but it's a IMO it's a pretty interesting data point.

maxfierke:
Thanks for throwing that together @dfreeman!

I think it's absolutely still something I'd be interested in seeing applied to the newer ember-concurrency core, and something I may look more into later (but someone else is also free to open an RFC if they're excited about it!) The big questions I would want answered are:

1. Can we apply to the [base `Task` class prototype](https://github.com/machty/ember-concurrency/blob/11548f93065030ac0a63abc879d247d0df5750f4/addon/-private/external/task/task.js#L20) somehow? (i.e. so all Tasks are executable like this, without needing to create a new function wrapper for `task.perform` and have to change the prototype to Task)
2. How supported is using `Object.setPrototypeOf` with `Function` in engines? Is it supported by spec or does it work accidentally?
3. What's the perf impact, if any? (e.g. Benchmarking against the latest v2)

If you or someone else is interested in tackling, I'd encourage you/them to open another RFC issue or PR, so that we can focus commentary on that specific approach/implementation.

PR #293, opened 2019-04-18
Revert Ember support

simonihmig:
As discussed in #290 this reverts the dropped support for older Ember versions and their test matrix in that PR to cover all the previously supported Ember versions, which were starting at 2.4. 

Although the Readme's badge (which was broken) stated support for Ember 1.13, but tests failed for that version. Didn't dig into it, but updated (fixed) the badge with Ember 2.4 as the minimum supported and tested version. Agree?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,11 @@ env:
   matrix:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
+    - EMBER_TRY_SCENARIO=ember-lts-2.4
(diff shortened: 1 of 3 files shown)


machty:
@simonihmig Thank you! So I should release this as 0.10.1?

simonihmig:
> So I should release this as 0.10.1?

@machty So technically this PR does not add anything that the published package would contain, as it only touched the test config and the Readme. But maybe it makes sense to publish it anyways, to at least document it in the Github release page / Changelog that Ember 2.4+ is still officially supported!? In that case 0.10.1 would be appropriate!

PR #299, opened 2019-06-03
chore(deps): upgrade ember-cli-sass & ember-cli-babel

buschtoens:
Upgrades these dependencies:

- `ember-cli-sass`: `^7.1.3` -> `^10.0.1`
  This allows running the dummy app with Node 12.
- `ember-cli-babel`: `^6.8.2` -> `^7.7.3`
  This enables the usage of decorators.
  Also removes the deprecated `babel.blacklist` config. This is now handled via the browser `targets.js`.

These were the only dependencies blocking my local development, but I would also recommend to upgrade the others soon:

image

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -26,7 +26,8 @@ module.exports = function(defaults) {
     babel: babelOptions,
 
     'ember-cli-babel': {
-      includePolyfill: process.env.EMBER_ENV === 'production'
+      includePolyfill:
+        process.env.EMBER_ENV === 'production' || Boolean(process.env.CI)
     },
 
     prember: {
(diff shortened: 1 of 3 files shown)


maxfierke:
@buschtoens mind rebasing this on the latest master? Might have some conflicts w/ the ember-cli-babel updates, but I think the sass updates to unblock Node 12 would still be good.

ember-cli-build.js
```diff
@@ -26,7 +26,8 @@ module.exports = function(defaults) {
     babel: babelOptions,
 
     'ember-cli-babel': {
-      includePolyfill: process.env.EMBER_ENV === 'production'
+      includePolyfill:
+        process.env.EMBER_ENV === 'production' || Boolean(process.env.CI)
```
buschtoens: You want me to drop this change?

maxfierke: Yeah, that'd be great! CI is using all evergreen browsers now, so the polyfill is probably not necessary these days.

PR #300, opened 2019-06-03
test: `reads('task.last.value')` tracked prop interop

buschtoens:
While upgrading our app to Ember Octane, I found that there is an issue with the interoperability of `task.last.value` and the `reads` computed property when tracked properties (`EMBER_METAL_TRACKED_PROPERTIES`) are enabled.

I don't yet understand the specific internals and haven't dug deeper, but I was able to conjure up a reliable minimal reproduction.

You can run the test case with:
`ember try:one ember-canary-tracked-properties --skip-cleanup --- ember s -e test`
The test module is named `Integration | reads('task.last.value')`.

This PR includes #299, because I couldn't get ember-cli to run otherwise.

### DIFF
--- a/tests/integration/reads-last-value-test.js
+++ b/tests/integration/reads-last-value-test.js
@@ -1,0 +1,45 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, settled } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import { waitForQueue, task } from 'ember-concurrency';
+import EmberObject from '@ember/object';
+import { reads } from '@ember/object/computed';
+
+module("Integration | reads('task.last.value')", function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('Ember Object Model', async function(assert) {
+    this.subject = EmberObject.extend({
+      task: task(function*() {
+        yield waitForQueue('afterRender');
+        return 1337;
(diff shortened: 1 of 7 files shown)


tests/integration/reads-last-value-test.js
```diff
@@ -0,0 +1,45 @@
...
+      lastValue: reads('task.last.value')
+    }).create();
+
+    await render(
+      hbs`reads: {{this.subject.lastValue}}; direct: {{this.subject.task.last.value}};`
+    );
+
+    assert.strictEqual(this.element.textContent.trim(), 'reads: ; direct: ;');
+
+    this.subject.otherTask.perform(); // toggle this line
```
buschtoens: The error _only_ occurs, if this line is executed. If you comment it out, the the test will pass.

tests/integration/reads-last-value-test.js
```diff
@@ -0,0 +1,45 @@
...
+        yield waitForQueue('afterRender');
+        return 1337;
+      }),
+
+      otherTask: task(function*() {
+        yield waitForQueue('afterRender');
+        return 1337;
+      }),
+
+      lastValue: reads('task.last.value')
```
buschtoens: If I substitute `last` with `lastSuccessful` instead, it also works.

buschtoens:
Fixed via https://github.com/emberjs/ember.js/pull/18073! :tada:

PR #302, opened 2019-06-17
Scheduler refactor / cleanup / e-c v2 groundwork

### LINKED ISSUE #175: task.last is the same as taskGroup.last
If task belongs to group, `last`, `lastRunning`, `lastPerformed`, `lastSuccessful`, `lastComplete`, `lastErrored`, `lastCanceled`, `lastIncomplete`, `performCount` properties would be properties of group, not of task. That means it's impossible to get a last value or last error of a particular task, if it belongs to group. It would be great to mention this in API documentation.

### LINKED ISSUE #278: task.cancelAll(); does not decremented numRunning
I call cancelAll() on task group, but numRunning  not decremented accordingly it does not recompute isRunning flag.
`  testTaskGroup: taskGroup().drop(),
    task: task(function*() {
      // for example
         while (true) {
                yield timeout(10000);
            }
    }).group('testTaskGroup'),
`
run this task 
this.testTaskGroup.isRunning == true
and call this.testTaskGroup.cancelAll()
but
this.testTaskGroup.isRunning == true
numRunning == 1

machty:
- Convert a lot of inner classes to ES6
- Use fewer computed properties and aliases in favor of simpler state POJOs
- Fix many bugs surrounding derived state and task groups
  - Closes #175 
  - Closes #278 
- Move all private addon files into `-private/...`
- Simplify the algorithm for computing derived state
  - Fewer loops and weird re-entrancy corner cases


addon/-private/utils.js
```diff
@@ -21,6 +21,10 @@ Arguments.prototype.resolve = function(value) {
   }
 };
 
+let guidId = 0;
+export function makeGuid() {
+  return `ec_${guidId++}`;
```
maxfierke: `__ec__private_guid_donttouchthiswhatareyoudoing_` 😏

machty:
@maxfierke @NullVoxPopuli @pzuraq (and please tag anyone else who might interested)

This is the PR that lays the groundwork for an ember-agnostic concurrency library, and enough people have been asking about this that I wanted to see if one (or more) of you were interested in championing the push to finish this PR and perhaps land ember-concurrency V2 along with the extracted concurrency lib. 

I simply don't have the time or work setup to put the requisite hours into making this happen, and I don't want this to languish, so if one of yall (or some yet untagged person) is interested, maybe we could do a kick-off orientation call as to the current state of this PR and what an ember-agnostic concurrency library would look like?

FWIW part of this morning's motivation is due to seeing this [Netflix talk](https://youtu.be/LtrXwX81CPE?t=1144) that discusses how they use e-c (though not mentioning the library's name for some reason) and how they might be working on their own React version soon. This PR includes a 99% feature complete concurrency lib and it'd feel like a missed opportunity for another one to pop up just because this languished. 

Honestly most of the hard work isn't so much the code, but it's questions of backwards compatibility and interop (things like, what happens if an app has e-c v1 / v2 dependencies; what happens if an EmberObject Task meets an ES6 Task class, etc) as well as what needs to happen to get tasks working well in Octane and Glimmer, and I just don't have the time / space in my brain for all those considerations.

maxfierke:
@machty I'm more than willing and interested in seeing this work shipped, so I would love to help shepherd this through with the help of folks who are more familiar with the interop constraints (thinking specifically about things like computed vs. tracked properties, interacting between the two etc. for derived state)

I think it would also be good to get @buschtoens involved, if he's got time and is willing, particularly for his take on implications for decorators and TypeScript.

addon/-private/ember-environment.js
```diff
@@ -0,0 +1,35 @@
...
+  assert(...args) {
+    assert(...args);
+  }
+
+  async(callback) {
+    join(() => schedule('actions', null, callback));
+  }
+
+  reportUncaughtRejection(error) {
+    setTimeout(function() {
```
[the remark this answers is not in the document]
sukima: Personally I think making a `nextTick` function would make more sense. And at this point `run.later` doesn't do much other then have a meaningful name and this case that name is misleading.

addon/-private/external/generator-state.js
```diff
@@ -0,0 +1,43 @@
...
+      return this.finalize(e, true);
+    } 
+  }
+
+  getIterator() {
+    if (!this.iterator && !this.done) {
+      this.iterator = this.generatorFactory();
+    }
+    return this.iterator;
+  }
```
sukima: Are these _iterators_ different then ES6 iterators (`Symbol.iterator`)?

addon/-private/external/scheduler/policies/execution-states.js
```diff
@@ -0,0 +1,9 @@
+// TODO: change to ints?
```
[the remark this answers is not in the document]
sukima: Oh I want to know this because I struggle with this all the time...

Why `int`s VS `string`s? What are the pros cons? This far I've found strings to offer better debugging for other developers and allows some consumers to avoid imports for comparing **if** they wanted to live on the edge (I know many devs who do dispite any arguments to the contrary, One of those "can't beat 'em join 'em" things. I use this pattern of constants all the time and I can never figure out if I should use ints or strings.

addon/-private/external/scheduler/policies/keep-latest-policy.js
```diff
@@ -0,0 +1,52 @@
+import BoundedPolicy from "./bounded-policy";
+import { STARTED, QUEUED, makeCancelState } from "./execution-states";
+
+const CANCELLED = makeCancelState(
+  `it belongs to a 'keepLatest' Task that was already running`
+);
```
sukima: oooh this is a neat pattern. 👍

addon/-private/external/scheduler/policies/execution-states.js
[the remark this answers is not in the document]
machty: Yeah I think I'd prefer to keep with strings.

addon/-private/external/generator-state.js
maxfierke: `this.iterator` is a generator function, so it would also be usable with `Symbol.iterator`. However, `GeneratorState` is not as-of-yet accessed this way and is more of a state wrapper/manager around an iterator than an iterator itself.

maxfierke:
Going to go ahead and merge this into the `v2` branch, now that it's up to date with `master` and is passing Travis. Development and further refinement will continue on the `v2` branch.

PR #310, opened 2019-07-23
Export `rawTimeout` from `ember-concurrency` module 

Turbo87:
`rawTimeout` was previously only available from `ember-concurrency/utils`. This PR promotes it to be available from the top-level module.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 import { computed } from '@ember/object';
-import { timeout, forever } from './utils';
+import { timeout, forever, rawTimeout } from './utils';
 import { Task, TaskProperty } from './-task-property';
 import { didCancel } from './-task-instance';
 import { TaskGroup, TaskGroupProperty } from './-task-group';
@@ -143,6 +143,7 @@ export {
   hash,
   race,
   timeout,
+  rawTimeout,
   waitForQueue,
   waitForEvent,
   waitForProperty,
--- a/addon/utils.js
+++ b/addon/utils.js
@@ -98,6 +98,11 @@ export const _ComputedProperty = ComputedProperty;
  * Yielding `timeout(ms)` will pause a task for the duration
  * of time passed in, in milliseconds.
  *
+ * This timeout will be scheduled on the Ember runloop, which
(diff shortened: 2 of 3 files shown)


Turbo87:
@machty any thoughts on this?

Turbo87:
@maxfierke do you think the CI failure is actually caused by the changes here? 🤔

maxfierke:
@Turbo87 Oop, sorry I missed this for 1.1.0. nope, I don't think so. Those buffering tests are a little flaky on CI (haven't narrowed the cause down yet.) Just restarted the failed one and I would expect it should pass.

Turbo87:
okay, thanks for the feedback! seems to be all green now :)

Turbo87:
> why someone might use this rather than `timeout`, though

my primary reason is that `rawTimeout()` (as opposed to `timeout()`) does not use Ember runloop timers, and so it does not block the test helpers that rely on `settled()` (like `click()`, `fillIn()`, etc.).

does that match your thoughts? 

the other thing I'm slightly worried about because I don't know enough about the details is: since this is *not* using the Ember runloop timers, do we need to do anything to sync properly with the runloop after using `yield rawTimeout()`? e.g. do we need something like `yield waitForQueue('something')` to avoid timing issues?

maxfierke:
> my primary reason is that `rawTimeout()` (as opposed to `timeout()`) does not use Ember runloop timers, and so it does not block the test helpers that rely on `settled()` (like `click()`, `fillIn()`, etc.).
> 
> does that match your thoughts?

yep, that makes sense.

> 
> the other thing I'm slightly worried about because I don't know enough about the details is: since this is _not_ using the Ember runloop timers, do we need to do anything to sync properly with the runloop after using `yield rawTimeout()`? e.g. do we need something like `yield waitForQueue('something')` to avoid timing issues?

The run-loop syncing already happens during the scheduling of the next generator step execution inside the task instance, so there shouldn't be any issues there. (should run in 'actions' queue)

Turbo87:
sweet, that's what I was hoping for. I guess with that information I should be able to come up with a bit of documentation :)

Turbo87:
@maxfierke finally had time again for this. let me know if this is sufficient as documentation :)

PR #314, opened 2019-08-13
Ensure waitForQueue throws on errors

mydea:
I ran into an issue, where `waitForQueue()` would never resolve. After some digging, I found out it was because I used `waitForQueue('sync')`, but the `sync` runloop was removed in 3.11. However, this was totally swallowed by `waitForQueue`, although e.g. `schedule('sync', func)` _did_ actually throw a descriptive error.

This PR fixes this by ensuring that an error thrown bubbles up correctly.

### DIFF
--- a/addon/-wait-for.js
+++ b/addon/-wait-for.js
@@ -6,7 +6,8 @@ import { isEventedObject, yieldableToPromise } from './utils';
 
 import {
   yieldableSymbol,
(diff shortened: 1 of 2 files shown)


maxfierke:
@mydea if you rebase off master, that should get this a passing build, and then I'd be happy to merge

mydea:
Done! :)

PR #316, opened 2019-08-15
Resolve `ember-maybe-import-regenerator` Conflict

jherdman:
Currently when `yarn install`ing in this repo the following warning will
be issued:

    warning ember-concurrency@1.0.0: "dependencies" has dependency "ember-maybe-import-regenerator" with range "^0.1.5" that collides with a dependency in "devDependencies" of the same name with version "^0.1.6"

I took a wild guess and assumed we should update the `"dependencies"`
version to match the most recent change.

Note that #315 should probably be merged first, and this rebased on top of it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "ember-disable-proxy-controllers": "^1.0.1",
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.0.0",
-    "ember-maybe-import-regenerator": "^0.1.6",
     "ember-notify": "^5.3.0",
     "ember-qunit": "4.1.2",
     "ember-resolver": "^4.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -71,7 +71,7 @@
     "babel-core": "^6.24.1",
     "ember-cli-babel": "^6.8.2",
     "ember-compatibility-helpers": "^1.2.0",
-    "ember-maybe-import-regenerator": "^0.1.5"
+    "ember-maybe-import-regenerator": "^0.1.6"
```
buschtoens: Listing a dependency in `dependencies` and `devDependencies` doesn't really make sense. The correct fix IMO, would be removing it from `devDependencies` and raising the version here in `dependencies` to ^0.1.6`.

jherdman: This is a great point. I wasn't really sure why it was duplicated. I'll try removing it from `devDependencies` to see what happens.

BTW are the failed tests normal? As far as I can tell they also pass sometimes 🕵

buschtoens: The failing tests seem to be flaky and not related to this PR.

jherdman:
Hey friends. Just wondering if there's anything else that needs to be done to get this merged.

PR #318, opened 2019-09-16
Ensure tasks are assigned to a container

maxfierke:
This fixes an issue with service injection inside
of encapsulated tasks on Ember 3.13+ by ensuring
tasks that are children of other tasks are
assigned a container to use for service injection.

### DIFF
--- a/tests/dummy/app/task-injection-test/controller.js
+++ b/tests/dummy/app/task-injection-test/controller.js
@@ -7,7 +7,7 @@ export default Controller.extend({
 
   myTask: task({
     fun: service(),
-    perform: function * () {
+    *perform() {
       let value = yield this.get('subtask').perform();
       return `${this.get('fun.foo')}-${value}`;
     },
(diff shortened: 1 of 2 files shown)


tests/dummy/app/task-injection-test/controller.js
```diff
@@ -7,18 +7,17 @@ export default Controller.extend({
 
   myTask: task({
     fun: service(),
-    perform: function * () {
+    *perform() {
```
Turbo87: seems unrelated?

Turbo87:
@maxfierke is the test suite failing without this patch? otherwise it might make sense to create a test for it that verifies that this works correctly with both old and new Ember versions

maxfierke:
@Turbo87 yeah, the build fails on master for `ember-beta` and `ember-canary` without this

tests/dummy/app/task-injection-test/controller.js
maxfierke: unrelated to the fix but _cleaner_. Didn't feel like it warranted omission 😛

PR #319, opened 2019-09-18
Add Type Definitions

buschtoens:
This copies the types from https://github.com/machty/ember-concurrency-decorators/pull/56 and together with that PR will finally enable type-safe usage of ember-concurrency.

Closes #209, though we should compare both.

Not 💯 sure what's left to do here and whether / how to test this.

/cc @jamescdavis @chriskrycho @dfreeman @josemarluedke

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,438 @@
+import EmberObject from '@ember/object';
+import {
+  UnwrapComputedPropertyGetter,
+  UnwrapComputedPropertyGetters
+} from '@ember/object/-private/types';
+
+import RSVP from 'rsvp';
+
+// Lifted from @types/ember__object/observable.d.ts
+interface Getter {
+  /**
+   * Retrieves the value of a property from the object.
+   */
+  get<K extends keyof this>(key: K): UnwrapComputedPropertyGetter<this[K]>;
+  /**
+   * To get the values of multiple properties at once, call `getProperties`
+   * with a list of strings or an array:
+   */
+  getProperties<K extends keyof this>(
+    list: K[]
+  ): Pick<UnwrapComputedPropertyGetters<this>, K>;
+  getProperties<K extends keyof this>(
+    ...list: K[]
+  ): Pick<UnwrapComputedPropertyGetters<this>, K>;
+}
+
+export type GeneratorFn<T, Args extends any[] = any[], R = any> = (
+  this: T,
+  ...args: Args
+) => IterableIterator<R>;
+
+export const all: typeof Promise.all;
+export const allSettled: typeof RSVP.allSettled;
+export const hash: typeof RSVP.hash;
+export const race: typeof Promise.race;
+
+export function timeout(ms: number): Promise<void>;
+
+/**
+ * Use `waitForEvent` to pause the task until an event is fired.
+ * The event can either be a jQuery event or an `Ember.Evented` event, or any
+ * event system where the object supports `.on()`, `.one()` and `.off()`.
+ */
+export function waitForEvent(
+  object: EmberObject | EventTarget,
+  eventName: string
+): Promise<Event>;
+
+/**
+ * Use `waitForProperty` to pause the task until a property on an object
+ * changes to some expected value.
+ *
+ * This can be used for a variety of use cases, including synchronizing with
+ * another task by waiting for it to become idle, or change state in some
+ * other way.
+ *
+ * If you omit the callback, `waitForProperty` will resume execution when the
+ * observed property becomes truthy. If you provide a callback, it'll be
+ * called immediately with the observed property's current value, and multiple
+ * times thereafter whenever the property changes, until you return a truthy
+ * value from the callback, or the current task is canceled.
+ *
+ * You can also pass in a non-`function` value in place of the callback, in
+ * which case the task will continue executing when the property's value
+ * becomes the value that you passed in.
+ */
+export function waitForProperty<T extends object, K extends keyof T>(
+  object: T,
+  key: K,
+  callbackOrValue?: (value: T[K]) => boolean | any
+): Promise<void>;
+
+/**
+ * Use `waitForQueue` to pause the task until a certain run loop queue is
+ * reached.
+ */
+export function waitForQueue(queueName: string): Promise<void>;
+
+export function task<Args extends any[], R>(
+  taskFn: GeneratorFn<unknown, Args, R>
+): Task<Args, Exclude<R, Promise<any>>>;
+export function task<Args extends any[], R>(encapsulatedTask: {
+  perform: GeneratorFn<unknown, Args, R>;
+}): Task<Args, Exclude<R, Promise<any>>>;
+
+export function taskGroup(): TaskGroupProperty;
+
+interface CommonTaskProperty {
+  restartable: () => TaskProperty;
+  drop: () => TaskProperty;
+  keepLatest: () => TaskProperty;
+  enqueue: () => TaskProperty;
+  maxConcurrency: (n: number) => TaskProperty;
+  cancelOn: (eventName: string) => TaskProperty;
+  group: (groupName: string) => TaskProperty;
+}
+
+export interface TaskProperty extends CommonTaskProperty {
+  evented: () => TaskProperty;
+  debug: () => TaskProperty;
+  on: (eventName: string) => TaskProperty;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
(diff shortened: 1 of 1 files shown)


chriskrycho:
I at least can give this a close look on Friday (have time allocated for miscellaneous open source woke then!).

index.d.ts
```diff
@@ -0,0 +1,438 @@
...
+   */
+  readonly error?: Error;
+
+  /**
+   * Cancels the task instance. Has no effect if the task instance has already
+   * been canceled or has already finished running.
+   */
+  cancel(): void;
+
+  catch(): RSVP.Promise<unknown>;
```
josemarluedke: Are there any downsides to using the native Promise here? 

```suggestion
  catch(): Promise<unknown>;
```

chriskrycho: Yes: it’s not what the library actually returns! RSVP is a compatible superset type-wise, so you can test it as a normal `Promise` in your own codebase as a consumer, but if you *need* to take advantage of its being RSVP, you can’t if you make this change.

index.d.ts
```diff
@@ -0,0 +1,438 @@
+import EmberObject from '@ember/object';
+import {
+  UnwrapComputedPropertyGetter,
+  UnwrapComputedPropertyGetters
+} from '@ember/object/-private/types';
```
mike-north: These private types change _all the time_, and are often in need of adjustment to respond to breaking TS compiler changes. _**Please do not consume these directly**_

index.d.ts
```diff
@@ -0,0 +1,438 @@
...
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface TaskGroupProperty extends CommonTaskProperty {}
+
+// Based on https://github.com/CenterForOpenScience/ember-osf-web/blob/7933316efae805e00723789809bdeb58a96a286a/types/ember-concurrency/index.d.ts
+
+/**
+ * Describes the state that the task instance is in.
+ * Can be used for debugging, or potentially driving some UI state.
+ */
+export enum TaskInstanceState {
```
mike-north: I'd avoid creating the `TaskInstance` and `TaskState` enums, since they suggest that _values_ (not just types) can be imported from [`ember-concurrency`](https://github.com/machty/ember-concurrency/blob/master/addon/index.js). If it's _necessary_ to make some utility types available to consumers, that should probably be done through a ghost module to avoid clashing with anything that might be exported as a value from the underlying library in the future

index.d.ts
```diff
@@ -0,0 +1,438 @@
...
+} from '@ember/object/-private/types';
+
+import RSVP from 'rsvp';
+
+// Lifted from @types/ember__object/observable.d.ts
+interface Getter {
+  /**
+   * Retrieves the value of a property from the object.
+   */
+  get<K extends keyof this>(key: K): UnwrapComputedPropertyGetter<this[K]>;
```
mike-north: This is going to break for cases like `this.get('foo.bar')`. Not all values that can be passed into this function are `keyof this`

index.d.ts
jamescdavis: Blame @chriskrycho 
https://gist.github.com/chriskrycho/9ed39076ccb698023e57e59080fedecd#file-ember-concurrency-d-ts-L4-L10
😜

mike-north: haha. I suspect the quality bar may be different different between Chris' personal gist and merging something in as official typings.

index.d.ts
mike-north: > Don't we just say getting in this way is not supported when using TS?

This is the current state of the types, given that we don't have the ability to use `unknown` as long as we continue to support TS 2.8.

Something like this (available in TS 3.0+) would support Ember's public API fully, without compromising type safety

```ts
interface Getter {
  get<K extends keyof this>(key: K): this[K];
  get(key: string): unknown;
}

interface Foo extends Getter {
  blue: string[];
  red: {
    green: string;
  }
}

const foo: Foo = {} as any;

foo.get('blue'); // string[];
foo.get('red.green'); // unknown
```

index.d.ts
```diff
@@ -0,0 +1,438 @@
...
+): Promise<void>;
+
+/**
+ * Use `waitForQueue` to pause the task until a certain run loop queue is
+ * reached.
+ */
+export function waitForQueue(queueName: string): Promise<void>;
+
+export function task<Args extends any[], R>(
+  taskFn: GeneratorFn<unknown, Args, R>
```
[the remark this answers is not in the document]
mike-north: ^ this is some great feedback, and another good reason to let tests (both for classic and octane use cases) be your guide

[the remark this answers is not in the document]
mike-north: > we just might need to communicate the caveat that it will break "classic" tasks and suggest the workaround (or another better one, if it exists).

I'm not sure I agree with this, given that vscode uses type information now _even for apps that do not have type-checking enabled_. One of the immediate wins of landing this would be having in-editor documentation/autocomplete around the `Task`, `TaskInstance`, `TaskGroup` APIs. If anything, "classic" usage may be significantly more important to cater to than octane usage at this time.

I still feel that we should cover both _as part of any initial release of type information to ember-concurrency_. See details in the previous PR here: https://github.com/machty/ember-concurrency/pull/209#issuecomment-403246551

index.d.ts
chriskrycho: That’s just crazy talk. :laughing: I agree here. Much as the enums are nice, if they can’t actually be imported we shouldn’t use them. Instead, we should just use literal types.

index.d.ts
chriskrycho: We should note that as a point for discussion Soon(tm). 3.0 was a Ling time ago and the official support policy we decided on currently only commits us to 3.5 and 3.6, I believe. We’ve intentionally supported a wider range, but it’s been quite some time and it’s worth revisiting what we support (outside this PR!).

index.d.ts
```diff
@@ -0,0 +1,438 @@
...
+}
+
+export type TaskInstance<T> = TaskInstanceBase<T> &
+  (
+    | InstanceDropped
+    | InstanceCanceled
+    | InstanceRunning
+    | InstanceWaiting
+    | InstanceError
+    | InstanceSuccess);
```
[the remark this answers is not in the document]
chriskrycho: It’s worth seeing if this works if we drop the magic bits imported for `get`—I haven’t had a chance to test, but I think it *may*, between that and extending from EmberObject? Because at that point it should be actually the same underlying type definition being invoked.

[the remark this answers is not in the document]
jamescdavis: Extending from `EmberObject` results in the same error. Removing the union of state interfaces works and also intersection with just one of the state interfaces works. Intersecting with the union of two or more of the states breaks. It looks like `UnwrapComputedPropertyGetter` can't handle unions of interfaces with different signatures, which makes sense given that it uses `infer`.

index.d.ts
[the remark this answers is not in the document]
jamescdavis: I agree, but also don't know how we'll make this work. As soon as you type `taskFn` as a callable (even `() => any`) you get the error about referencing the class in its own base expression, because, well, you are. When it's typed as just `any` (what we've been doing in our app) I'm guessing the compiler just ignores it outside the function. There may be a way to type this so that it will work, but I'm not seeing it.

index.d.ts
```diff
@@ -0,0 +1,438 @@
...
+export function task<Args extends any[], R>(
+  taskFn: GeneratorFn<unknown, Args, R>
+): Task<Args, Exclude<R, Promise<any>>>;
+export function task<Args extends any[], R>(encapsulatedTask: {
+  perform: GeneratorFn<unknown, Args, R>;
+}): Task<Args, Exclude<R, Promise<any>>>;
+
+export function taskGroup(): TaskGroupProperty;
+
+interface CommonTaskProperty {
```
NullVoxPopuli: should this also have `.withTestWaiter`? or is it recommended that people use the wrapping version?

NullVoxPopuli:
I think this method should go somewhere:

```ts
import Task from 'ember-concurrency/task';

type ECTask<Args extends Array<any>, Return> = (
  ...args: Args
) => Generator<any /* potentially yielded types */, Return, unknown>;

export function taskFor<Args extends any[], Return = void>(generatorFn: ECTask<Args, Return>) {
  return (generatorFn as any) as Task<Args, Return>;
}

```

image

maxfierke:
Thanks so much for your work on this! However, we're closing this in favor of #357, which has been merged and will be part of 1.2.0

PR #321, opened 2019-10-07
Use ember get in case object isn't an EmberObject

jrjohnson:
When using a TaskGroup on an octane glimmer component `get` does not
exist on the component. Ember.get shields us from this and just works.

Fixes machty/ember-concurrency-decorators#57


buschtoens:
Thanks for fixing this! 👏

jrjohnson:
I very much wanted to add tests @maxfierke, but was daunted by it since `EmberObject` is pretty deeply embedded in the current tests and in things like `Task`. Finally decided to just go with the obvious quick fix especially given other work like #302 that seems to address this issue head on.

PR #324, opened 2019-10-13
Use ember get for ES Classes

jrjohnson:
I missed some of these in #321. Though only the one in `-task-state-mixin.js ` seems to be an issue for my application, I didn't see the harm in protecting the others as well.

I did a walk through of all other instances of `.get()` not on `this` and these were the last ones I could find.

### DIFF
--- a/addon/-scheduler.js
+++ b/addon/-scheduler.js
@@ -129,12 +129,12 @@ function flushTaskCounts(tasks) {
 function updateTaskChainCounts(task) {
   let numRunning = task.numRunning;
   let numQueued  = task.numQueued;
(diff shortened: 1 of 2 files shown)


jrjohnson:
@maxfierke Checking in to see how tests are coming. Hoping for a release with this fix, if I can help with tests I'm happy to I just need a blueprint on how you want to test both scenarios.

maxfierke:
@jrjohnson thanks for the ping. I've got a PR up here: https://github.com/machty/ember-concurrency/pull/325 if you want to take a quick look and let me know if there's any Glimmer component or ES class cases you can think of that might be missing?

Issue #329, opened 2019-11-19
`rawTimeout()` does not work with `race()`


Turbo87:
```js
let result = yield race([someOtherPromise, rawTimeout(5000)]);
```

I've tried the above code, but it seems like the `race()` function is not considering the return value of `rawTimeout()` as something that can be waited upon. Using `timeout()` works fine though, so now I'm wondering if we should adjust the implementation of `rawTimeout()` to roughly match the one from `timeout()` 🤔 

/cc @maxfierke @machty

maxfierke:
@Turbo87 good catch, I think you're right:

https://github.com/machty/ember-concurrency/blob/3d96d8490c13220b23f105e3d535ccbd55d7534c/addon/-cancelable-promise-helpers.js#L123-L126

The cancelable promise helpers only seem to take into account those too. I think there's a built-in assumption that the cancelable promise helpers would be passed promise-ified task instances (which always have an `__ec_cancel__`) or TaskInstances, but that doesn't hold true for the more loosely defined "yieldables" that folks would probably expect to work here.

This is probably what's happening in #308 too.

Changing the implementation to make it similar to `timeout` so that it has an `__ec_cancel__` is a quick fix for `rawTimeout`, but there's probably some other refactoring that needs to happen to work with the `WaitFor`-derived yieldables too. I'll take a look at this tonight, but if you get a PR up for the `rawTimeout` stuff before then, I'd merge that.

maxfierke:
Closed by #331

PR #331, opened 2019-11-20
Refactor yieldable cancelation to work with cancelable promise helpers

### LINKED ISSUE #308: `waitForEvent` handlers inside of `race` are not always cleaned up
We have discovered an issue where event handlers added by `waitForEvent` are not cleaned up, if they occur within `race`:

```
      yield race([
        waitForEvent(window, 'mouseup'),
        waitForEvent(window, 'touchend'),
      ]);
```

When `yield`ing directly to `waitForEvent`, the handlers are removed when expected (e.g. when the component is destroyed).

As a workaround, it is possible to wrap individual events in their own tasks, and `race` on those. In that case, the event handlers are cleaned up as expected.

```
      yield race([
        this.wrappedMouseUp.perform(),
        this.wrappedTouchEnd.perform(),
      ]);
```

A reduced test case showing the issue and the workaround: [e-c-waitForEvent.zip](https://github.com/machty/ember-concurrency/files/3394365/e-c-waitForEvent.zip)

maxfierke:
The refactor addresses some gaps between yieldable implementations
around how cancelation is handled, especially when used with cancelable promise helpers.

Resolves #308, #329

### DIFF
--- a/addon/utils.js
+++ b/addon/utils.js
@@ -1,5 +1,5 @@
 import { later, cancel } from '@ember/runloop';
-import { Promise, defer } from 'rsvp';
+import { defer } from 'rsvp';
 import ComputedProperty from '@ember/object/computed';
 import Ember from 'ember';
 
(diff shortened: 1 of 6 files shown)


addon/utils.js
```diff
@@ -121,14 +149,7 @@ export const _ComputedProperty = ComputedProperty;
...
 export function timeout(ms) {
-  let timerId;
-  let promise = new Promise(r => {
-    timerId = later(r, ms);
-  });
-  promise.__ec_cancel__ = () => {
-    cancel(timerId);
-  };
-  return promise;
+  return new TimeoutYieldable(ms);
```
Turbo87: correct me if I'm wrong, but this change means it no longer works with async/await, right? maybe we could add a then() method that uses the yieldable method?

maxfierke: the base class `Yieldable` has a `then` defined on it that will turn the yieldable into a cancelable promise, so it _should_ still work with async/await. I noticed we don't have any tests for that, so I'm going to add those once I get back to this tonight.

maxfierke:
Noticed some stuff this morning that I should clean up, so I'm holding off on merging for now.

addon/utils.js
Turbo87: 🤦‍♂ sorry for the noise. you're totally right!

Issue #271, opened 2019-01-16
Does not work with native classes


mfeckie:
Using Ember 3.6, ember-concurrency 0.8.26

Given a component defined with 'classic' syntax, all is good.

```js
import Component from '@ember/component'
import { task, timeout } from 'ember-concurrency';

export default Component.extend({
  result: null,
  doStuff: task(function*(){
    yield timeout(1000);
    this.set('result', 'done');
  })
});
```

If, however, we use native class syntax

```js
import Component from '@ember/component';
import { task, timeout } from 'ember-concurrency';

export default class NativeTaskComponent extends Component {
  result = null;
  doStuff = task(function*() {
    yield timeout(1000);
    this.set('result', new Date());
  });
}
```

We see `-task-property.js:620 Uncaught Error: It looks like you tried to perform a task via this.nameOfTask.perform(), which isn't supported. Use 'this.get('nameOfTask').perform()' instead.`

I've spent a little time trying to figure out why this is the case and have found that when creating a TaskProperty, the function passed to `super` does not get executed https://github.com/machty/ember-concurrency/blob/master/addon/-task-property.js#L416-L437

```js
export class TaskProperty extends _ComputedProperty {
  constructor(taskFn) {
    let tp;
    console.log('called before super');
    super(function(_propertyName) {
      console.log('called inside super anonymous function');
      taskFn.displayName = `${_propertyName} (task)`;
...
```

The transpiled output of the classic invocation is

```js
define("dummy/components/classic-task/component", ["exports", "ember-concurrency"], function (_exports, _emberConcurrency) {
  "use strict";

  Object.defineProperty(_exports, "__esModule", {
    value: true
  });
  _exports.default = void 0;

  var _default = Ember.Component.extend({
    result: null,
    doStuff: (0, _emberConcurrency.task)(function* () {
      yield (0, _emberConcurrency.timeout)(1000);
      this.set('result', 'done');
    })
  });

  _exports.default = _default;
});
```

Whilst the native version is

```js
define("dummy/components/native-task/component", ["exports", "ember-concurrency"], function (_exports, _emberConcurrency) {
  "use strict";

  Object.defineProperty(_exports, "__esModule", {
    value: true
  });
  _exports.default = void 0;

  function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }

[...]

PR #334, opened 2019-11-25
Add some brief documentation on use with Octane/decorators

maxfierke:
Raised in issue #271, there is currently no official documentation
on using ember-concurrency with native classes and other Octane-related
use-cases.

This adds a brief section to "Task Function Syntax" in the docs that
shows an example of use with a Glimmer component and mentions the
ember-concurrency-decorators addon as a more concise approach to the
default decorators via the task() property.

Thanks @jenweber for the Glimmer example!


jenweber:
This is super, thanks for following through to add this!

PR #338, opened 2019-12-13
Avoid using incrementProperty to workaround duplicate auto-tracking assertion in Ember 3.15

### LINKED ISSUE #337: Issue with `performCount` on Ember v3.15
We tried to update our app to Ember 3.15 this morning and the test suite of one of our apps uncovered the following issue:

> Assertion Failed: You attempted to update `performCount` on `<(unknown):ember22520>`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.
>
> `performCount` was first used:
>
> - While rendering:
>     (instance of a `(unknown function)` modifier)
> 
> Stack trace for the first usage: 
>     at get (http://localhost:7357/assets/vendor.js:39430:11)
>     at Class.incrementProperty (http://localhost:7357/assets/vendor.js:53987:72)
>     at Class.schedule (http://localhost:7357/assets/vendor.js:158913:12)
>     at Class._performShared (http://localhost:7357/assets/vendor.js:160316:23)
>     at Class.perform (http://localhost:7357/assets/vendor.js:160286:19)
>     at XValidatedInput._focus (http://localhost:7357/assets/vendor.js:143177:33)
>     at XValidatedInput.focusIn (http://localhost:7357/assets/vendor.js:143157:12)
>     at XValidatedInput.trigger (http://localhost:7357/assets/vendor.js:59364:23)

I haven't been able to come up with a reproduction yet, but I'll update this issue if I can provide one.

maxfierke:
(Hopefully) Fixes #337 🤞 

Draft until there's a reproducible test case

### DIFF
--- a/addon/-scheduler.js
+++ b/addon/-scheduler.js
@@ -38,7 +38,7 @@ const Scheduler = EmberObject.extend({
         // This tracking logic is kinda spread all over the place...
         // maybe TaskInstances themselves could notify
         // some delegate of queued state changes upon cancelation?
-        taskInstance.task.decrementProperty('numQueued');
+        set(taskInstance.task, 'numQueued', taskInstance.task.numQueued - 1);
       }
 
       taskInstance.cancel(cancelReason);
@@ -51,8 +51,8 @@ const Scheduler = EmberObject.extend({
 
   schedule(taskInstance) {
     set(this, 'lastPerformed', taskInstance);
-    this.incrementProperty('performCount');
-    taskInstance.task.incrementProperty('numQueued');
+    set(this, 'performCount', this.performCount + 1);
+    set(taskInstance.task, 'numQueued', taskInstance.task.numQueued + 1);
     this.queuedTaskInstances.push(taskInstance);
     this._flushQueues();
   },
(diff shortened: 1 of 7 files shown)


Turbo87:
@maxfierke yep, I ran it through our CI and it indeed seems to fix the issue. This is slightly surprising to me though, since I would expect `incrementProperty()` to have the exact same implementation as what you replaced it with 😅

pichfl:
This is the implementation: 
https://github.com/emberjs/ember.js/blob/d4dc4b4cc587cbacdfb1958fbd03eb9770efe9ab/packages/%40ember/-internals/runtime/lib/mixins/observable.js#L439-L445

My wild guess is, that this is caused by some of the magic behind the nested getters in tracked properties (since nested props are incremented here) and setting them directly might work around that. 

In other words: Works in my project as well.

patocallaghan:
This fix worked for us too 👍

maxfierke:
@Turbo87 @pichfl @patocallaghan would any of you be able to provide a small test case or steps for reproduction for this issue? I'd like to have a test in place before merging this, and I'm not sure how to reproduce this, as the test suite is passing against 3.15+

Turbo87:
I finally was able to come up with a small repro:

```js
    this.owner.register(
      'component:e-c-test',
      Ember.Component.extend({
        layout: hbs`<input>`,

        focusIn() {
          this.exampleTask.perform();
        },

        exampleTask: task(function*() {
          yield timeout(100);
        }),
      })
    );

    await render(hbs`<ECTest {{autofocus}} />`);
```

with `{{autofocus}}` coming from https://github.com/qonto/ember-autofocus-modifier, but I suspect with a `did-insert` it would probably also trigger.

maxfierke:
Spent a little time playing around with `did-insert` before finding that auto-tracking is disabled for the `@ember/render-modifiers`, so they didn't trigger the issue (see https://github.com/emberjs/ember-render-modifiers/commit/2a8d1a10fb4c7e80e01ae86bb1a012e72632994a). Just went with the `{{autofocus}}` example and was able to reproduce in a test. Thanks a bunch @Turbo87! I'll get this merged and released sometime in the next day or so.

PR #344, opened 2020-02-06
Mark state on TaskInstance as tracked

### LINKED ISSUE #343: Auto-tracked getter fails to recompute when yielding
I’ve been struggling to figure out how to word this issue, so here’s a repro that I think demonstrates the problem:

http://jgwhite.co.uk/ember-concurrency-auto-tracking-repro/

Click the two “Run” buttons and you’ll see that when the task yields, auto-tracking seems to somehow break down.

jgwhite:
This fixes #343 

**TODO:**

- [ ] Get the tests passing
- [ ] Try to summarize what’s going on here!
- [ ] Write some tests

### DIFF
--- a/addon/-task-instance.js
+++ b/addon/-task-instance.js
@@ -13,6 +13,7 @@ import {
   YIELDABLE_CANCEL,
   RawValue
 } from './utils';
+import { tracked } from '@glimmer/tracking';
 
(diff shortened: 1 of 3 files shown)


jgwhite:
I spoke to @pzuraq about this and it seems like it’s on the right track (pun intended). Will work on it some more as soon as possible.

PR #355, opened 2020-04-29
add types definitions that also support ember-concurrency-decorators

NullVoxPopuli:
After initial review
- [ ] review #319, and address feedback from that PR
- [ ] add some tests
  - [ ] Octane App with ember-concurrency-decorators
  - [ ] Classic App with classic function wrappers
- [ ] de-namespace, as we can just export straight from this file, as TS already handles the namespacing because the index.d.ts will be in node-modules

------------------------------------------------------------------

Huge thanks to @buschtoens for the work done in https://github.com/machty/ember-concurrency/pull/319

these types enable the following, given these helper functions:
```ts
import Task from 'ember-concurrency/task';

type ECTask<Args extends Array<any>, Return> = (
  ...args: Args
) => Generator<any /* potentially yielded types */, Return, unknown>;

export function taskFor<Args extends any[], Return = void>(generatorFn: ECTask<Args, Return>) {
  return (generatorFn as any) as Task<Args, Return>;
}
```

we can have inferred types from ember-concurrency-decorators!
image

image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,249 @@
+declare module 'ember-concurrency' {
+  import Task from 'ember-concurrency/task';
+
+  type TaskDecorator = PropertyDecorator & {
+    drop(): TaskDecorator;
+    keepLatest(): TaskDecorator;
+    restartable(): TaskDecorator;
+    maxConcurrency(num: number): TaskDecorator;
+    enqueue(): TaskDecorator;
+    withTestWaiter(): PropertyDecorator;
+  };
+
+  export function task(
+    generator: Function
+  ): PropertyDecorator & {
+    drop(): TaskDecorator;
+    keepLatest(): TaskDecorator;
+    restartable(): TaskDecorator;
+    maxConcurrency(num: number): TaskDecorator;
+    enqueue(): TaskDecorator;
+    withTestWaiter(): PropertyDecorator;
+  };
+  export function timeout(wait: number): Promise<void>;
+
+  export function didCancel(error: Error): boolean;
+}
+
+declare module 'ember-concurrency/task' {
+  import TaskInstance from 'ember-concurrency/task-instance';
+
+  export enum TaskState {
+    RUNNING = 'running',
+    QUEUED = 'queued',
+    IDLE = 'idle',
+  }
+
+  /**
+   * The `Task` object lives on a host Ember object (e.g.
+   * a Component, Route, or Controller). You call the
+   * `.perform()` method on this object
+   * to create run individual `TaskInstance`s,
+   * and at any point, you can call the `.cancelAll()`
+   * method on this object to cancel all running or enqueued
+   * `TaskInstance`s.
+   */
+  export default class Task<PerformArgs extends any[] = any[], PerformReturn = void> {
+    /**
+     * `true` if any current task instances are running.
+     */
+    readonly isRunning: boolean;
+
+    /**
+     * `true` if any future task instances are queued.
+     */
+    readonly isQueued: boolean;
+
+    /**
+     * `true` if the task is not in the running or queued state.
+     */
+    readonly isIdle: boolean;
+
+    /**
+     * The current state of the task: `"running"`, `"queued"` or `"idle"`.
+     */
+    readonly state: TaskState;
+
+    /**
+     * The most recently started task instance.
+     */
+    readonly last: TaskInstance<PerformArgs, PerformReturn>;
+
+    /**
+     * The most recent task instance that is currently running.
+     */
+    readonly lastRunning: TaskInstance<PerformArgs, PerformReturn>;
+
+    /**
+     * The most recently performed task instance.
+     */
+    readonly lastPerformed: TaskInstance<PerformArgs, PerformReturn>;
+
+    /**
+     * The most recent task instance that succeeded.
+     */
+    readonly lastSuccessful: TaskInstance<PerformArgs, PerformReturn>;
+
+    /**
+     * The most recently completed task instance.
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -0,0 +1,249 @@
+declare module 'ember-concurrency' {
+  import Task from 'ember-concurrency/task';
+
+  type TaskDecorator = PropertyDecorator & {
+    drop(): TaskDecorator;
+    keepLatest(): TaskDecorator;
+    restartable(): TaskDecorator;
+    maxConcurrency(num: number): TaskDecorator;
+    enqueue(): TaskDecorator;
+    withTestWaiter(): PropertyDecorator;
```
maxfierke, line 10: this is defined by `ember-test-waiters`. perhaps this should live there? (is that a possible thing?)

index.d.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  export function task(
+    generator: Function
+  ): PropertyDecorator & {
+    drop(): TaskDecorator;
+    keepLatest(): TaskDecorator;
+    restartable(): TaskDecorator;
+    maxConcurrency(num: number): TaskDecorator;
+    enqueue(): TaskDecorator;
+    withTestWaiter(): PropertyDecorator;
```
maxfierke, line 21: same q

index.d.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+
+declare module 'ember-concurrency/task' {
+  import TaskInstance from 'ember-concurrency/task-instance';
+
+  export enum TaskState {
+    RUNNING = 'running',
+    QUEUED = 'queued',
+    IDLE = 'idle',
+  }
```
maxfierke, line 35: My understanding based on some feedback in #319, is that we should not export this in a type definition, since there is no such enum available to import.

index.d.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    /** task instance ran to completion (even if an exception was thrown) */
+    FINISHED = 'finished',
+
+    /** task instance is currently running (returns true even if is paused on a yielded promise) */
+    RUNNING = 'running',
+
+    /** task instance hasn't begun running yet (usually because the task is using the `TaskProperty#enqueue` task modifier) */
+    WAITING = 'waiting',
+  }
```
maxfierke, line 154: same deal here, re: exporting of an enum

index.d.ts
```diff
@@ -0,0 +1,249 @@
...
+    drop(): TaskDecorator;
+    keepLatest(): TaskDecorator;
+    restartable(): TaskDecorator;
+    maxConcurrency(num: number): TaskDecorator;
+    enqueue(): TaskDecorator;
+    withTestWaiter(): PropertyDecorator;
+  };
+
+  export function task(
+    generator: Function
```
maxfierke, line 14: should this be a `GeneratorFunction` (or w/e the TypeScript-equivalent type name is)?

index.d.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * The current state of the task: `"running"`, `"queued"` or `"idle"`.
+     */
+    readonly state: TaskState;
+
+    /**
+     * The most recently started task instance.
+     */
+    readonly last: TaskInstance<PerformArgs, PerformReturn>;
```
maxfierke, line 70: this (and the other `last*` properties) can also be `null`. Do these need to be denoted as nullable?

index.d.ts
NullVoxPopuli: maybe? but should https://github.com/bendemboski/ember-concurrency-test-waiter get pulled in to this addon since it's a useful tool to help in testing with tasks?

index.d.ts
NullVoxPopuli: yeah, these probably don't need to be exported -- this is _mostly_ copy paste from long ago

index.d.ts
NullVoxPopuli: depends on if any type inference could be gained from doing so, like how I'm getting inference with decorators

index.d.ts
NullVoxPopuli: oh! yes, you're absolutely right. (though is it null or undefined?)

NullVoxPopuli:
> Could you provide some details on what's different here versus #319?

the main difference is that I've added some generics that help with type inference with decorators (see screenshots in original comment)

> Does it address feedback raised there?

nope, I need to look at that

> Are there TypeScript version considerations?

no clue, I'm only using 3.7.x in my apps

> Is there a way to test/validate these (something that could be thrown in CI)?

yea - we could build this in to the repo with full ember apps that aren't the dummy app, and try to use all of the APIs and make sure there are no type errors?

index.d.ts
```diff
@@ -0,0 +1,249 @@
...
+    drop(): TaskDecorator;
+    keepLatest(): TaskDecorator;
+    restartable(): TaskDecorator;
+    maxConcurrency(num: number): TaskDecorator;
+    enqueue(): TaskDecorator;
+    withTestWaiter(): PropertyDecorator;
+  };
+  export function timeout(wait: number): Promise<void>;
+
+  export function didCancel(error: Error): boolean;
```
chancancode, line 25: This should probably be `export function didCancel(reason: unknown): reason is Error;`. In a promise callback, I don't think you necessarily know the reject reason object is an `Error`, and so if you require this parameter to be `Error` TypeScript might not let you pass it.

maxfierke:
Thanks so much for your work on this! However, we're closing this in favor of #357, which has been merged and will be part of 1.2.0

PR #357, opened 2020-06-02
Add type definition based on API docs

chancancode:
Yet another attempt at adding types.

Quite similar to #355, but this includes every item (and only those items) in the public API documentation. Also uses interfaces over classes, since the e-c classes are not meant to be instantiated manually or subclassed by apps.


addon/index.d.ts
```diff
@@ -0,0 +1,759 @@
...
+  // readonly last: TaskInstance<T> | null;
+  // readonly lastRunning: TaskInstance<T> | null;
+  // readonly lastPerformed: TaskInstance<T> | null;
+  // readonly lastSuccessful: TaskInstance<T> | null;
+  // readonly lastComplete: TaskInstance<T> | null;
+  // readonly lastErrored: TaskInstance<T> | null;
+  // readonly lastCanceled: TaskInstance<T> | null;
+  // readonly lastIncomplete: TaskInstance<T> | null;
+  // readonly performCount: number;
+  // cancelAll(options?: { reason?: string, resetState?: boolean }): void;
```
chancancode, line 142: I believe these are supposed to be public. However, they are not documented on the public API documentation, so to cautions, I left them out. If they become documented in the future we can add them.

addon/index.d.ts
```diff
@@ -0,0 +1,759 @@
...
+  /**
+   * Cancels the task instance. Has no effect if the task instance has
+   * already been canceled or has already finished running.
+   *
+   * @param cancelReason Defaults to `".cancel() was explicitly called"`.
+   */
+  cancel(cancelReason?: string): void;
+
+  /**
+   * Returns a promise that resolves with the value returned
```
chancancode, line 237: I think this is wrong, or at least worded confusingly. I kept it because I am trying to copy the docs 1:1.

addon/index.d.ts
```diff
@@ -0,0 +1,759 @@
...
+   * from the task's (generator) function, or rejects with
+   * either the exception thrown from the task function, or
+   * an error with a `.name` property with value `"TaskCancelation"`.
+   */
+  then<TResult1 = T, TResult2 = never>(
+    onfulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | null,
+    onrejected?: ((reason: any) => TResult2 | PromiseLike<TResult2>) | null
+  ): Promise<TResult1 | TResult2>;
+
+  catch<TResult = never>(
```
chancancode, line 247: This doesn't have docs?

addon/index.d.ts
```diff
@@ -0,0 +1,759 @@
...
+    onrejected?: ((reason: any) => TResult | PromiseLike<TResult>) | null
+  ): Promise<T | TResult>;
+
+  finally(onfinally?: (() => void) | null): Promise<T>;
```
chancancode, line 251: This doesn't have docs?

addon/index.d.ts
```diff
@@ -0,0 +1,759 @@
...
+ *
+ * See [Managing Task Concurrency](/#/docs/task-concurrency) for an
+ * overview of all the different task modifiers you can use and how
+ * they impact automatic cancelation / enqueueing of task instances.
+ */
+export interface TaskProperty<T = unknown, Args extends any[] = []> extends ComputedProperty<Task<T, Args>> {
+  volatile: never;
+  readOnly: never;
+  property: never;
+  meta: never;
```
chancancode, line 267: This extends `ComputedProperty<Task<T, Args>>` because that seems to be the only public way to make `.get` work. However, the CP properties doesn't actually exist on a `TaskProperty` (not on recent Ember versions anyway), hence retyping to `never`.

addon/index.d.ts
machty: I can confirm that they should all be documented/public.

machty:
This lgtm to I'm behind on the TypeScript discussion / tradeoffs when it comes to e-c so I'll defer to others.

addon/index.d.ts
chancancode: I have added (mostly just copied) the docs.

chancancode:
@maxfierke I have added (mostly just copied) the docs for `TaskGroup`

addon/index.d.ts
chancancode: Should I add docs for this or remove it?

addon/index.d.ts
chancancode: Should I add docs for this or remove it?

addon/index.d.ts
maxfierke: Ahh, good catch. Yes this should be documented API.

addon/index.d.ts
maxfierke: Yep, this should also be documented API. Thanks!

addon/index.d.ts
maxfierke: I think keeping it the same is probably fine for now. I'm planning to do a broader revisiting of the docs shortly.

addon/index.d.ts
chancancode: Actually, they are already "documented", in that the exists and are marked as public, just didn't have a text description.

addon/index.d.ts
chancancode: Actually, they are already "documented", in that the exists and are marked as public, just didn't have a text description.

maxfierke:
@chancancode I think this looks great! Before I approve, are there any caveats you can think of that would be worth mentioning or potentially breaking for folks (outside of people using their own types)? I'm thinking specifically about TypeScript not supporting changes to the return types of generators at one point, IIRC.

chancancode:
> Before I approve, are there any caveats you can think of that would be worth mentioning or potentially breaking for folks (outside of people using their own types)?

These PR adds the following top level **type** symbols:

* `TaskGenerator`
* `TaskFunction`
* `TaskFunctionArgs`
* `TaskFunctionReturnType`
* `Task`
* `TaskGroup`
* `TaskInstance`
* `TaskProperty`
* `TaskCancelation`
* `Yieldable`
* `Evented`

These can be imported from the `ember-concurrency` module by name when using TypeScript, so they are effectively public API once released.

Because they are type symbols, non-TypeScript projects won't be able to import them, and TypeScript projects won't be able to do anything with them other than using them as types (e.g. you cannot do `new Task()` because that would require `Task` being a "value" symbol, but `Task` is just a type here (which is erased from the code once compiled).

These names are effectively "reserved". It would be okay if, for example, ember-concurrency wants to export the same, currently internal, `Task` class as a public API for subclassing. In that case, we will just update the TypeScript definition from an interface to a class, all the existing properties will match up just fine. However, if ember-concurrency wants to use the `Task` named export for something completely different (say, `export cons Task = "Task, the string.";`), that would not be okay since the type definition will be forced to find a different name for what is currently exported as `Task`, and that would be a breaking change.

In practice, I think this seems unlikely to be a problem since I was quite careful to pick the names to match the defacto public names in the documentation (`@class Task`, etc). It's also possible to reclaim these symbols for something else by bumping major version.

> I'm thinking specifically about TypeScript not supporting changes to the return types of generators at one point, IIRC.

I think this is referring to the fact that the _decorator_ cannot change the type of the property it's decorating. So since most people write `@task *foo()...` you cannot convince typescript to let you call `this.foo.perform(...)` no matter what.

This is not really an ember-concurrency problem since the decorators are outside of this repository. Based on the way I typed it, the classic style `foo: task(function *() { ... })` should work _when used with `this.get('foo').perform(...)`. I think `this.foo.perform(...)` still won't work there since the task function is typed to return a `TaskProperty`.

The workaround in our project (very similar to @NullVoxPopuli's) is the following utility functions:

```ts
// app/types/ember-concurrency.d.ts

import {
  Task,
  TaskFunction,
  TaskFunctionArgs,
  TaskFunctionReturnType,
  TaskInstance
} from 'ember-concurrency';

/**
 * No-op typecast function that turns what TypeScript believes to be a
 * generator function into a Task.
 *
 * ```js
 * import { taskFor } from 'direwolf/types/ember-concurrency';
 *
 * class Foo extends EmberObject {
 *   @task *myTask() {
 *     // ...
 *   }
 *
 *   someMethod() {
 *     this.myTask.perform(); // TypeError
 *     taskFor(this.myTask).perform(); // ok!
 *   }
 * }
 * ```
 *
 * @param task The task. Note that this is purely a typecast function,
 *   it does not in affect accept a task generator function as input.
 */
export function taskFor<T extends TaskFunction<any, any[]>>(task: T):
  Task<TaskFunctionReturnType<T>, TaskFunctionArgs<T>>;
```

```js
// app/types/ember-concurrency.js

import { assert } from '@ember/debug';

export function taskFor(task) {
  assert(
    `${task} does not appear to be a task!`,
    task && typeof task.perform === 'function'
  );

  return task;
}
```

In actual usage:


```js
// @ts-check
// app/components/foo.js

import { action } from '@ember/object';
import Component from '@glimmer/component';

import { task } from 'ember-concurrency-decorators';

import { perform, taskFor } from 'my-app/types/ember-concurrency';

export default class extends Component {
  /**
   * @param {string} foo
   */
  @task *myTask(foo) {
    // ...
    return 1;
  }

  async someMethod() {
    this.myTask.perform("foo"); // error
    this.myTask.isRunning; // error

    taskFor(this.myTask).perform("foo"); // ok
    taskFor(this.myTask).perform(); // error
    taskFor(this.myTask).perform(123); // error

    let foo: number = await taskFor(this.myTask).perform("foo"); // ok
    let bar: string = await taskFor(this.myTask).perform("foo"); // error

    taskFor(this.myTask).isRunning; // ok

    let task = taskFor(this.myTask);

    task.perform("foo"); // ok
    task.perform(); // error
    task.perform(123); // error

    foo = await task.perform("foo"); // ok
    bar = await task.perform("foo"); // error

    task.isRunning; // ok
  }
}
```

chancancode:
> would you be willing to add a page to the docs w/ that example?

Sure, but I'll need a couple of pointers:

* Is there an inline way to write code snippets, or do I just make the file in the `snippets` folder?
* Does the "snippets" support `.ts` or `.d.ts`?
* Should I be using decorators in the example or using classic syntax? I suppose the latter could work, but it doesn't work super well with TypeScript, but perhaps it works good enough for the purpose of the example.
* Glimmer or classic component?
* For classic syntax at least, I believe `.get` is supposed to Just Works™ here, should I be using that?

> would it make sense to include such a type-cast function as part of this?

I'm not sure. For one, it's kind of weird to have an empty function that doesn't do anything for non-TypeScript users. The other is that it's basically an "unsafe" function that works by lying to the compiler to mask the type error. I don't think it's particular dangerous, because at the end of the day it's just a weird quirk/fallout, but I'm not sure it's good to include it as an official API.

maxfierke:
> * Is there an inline way to write code snippets, or do I just make the file in the `snippets` folder?

Yep, you can do this in HBS:

```hbs
{{! BEGIN-SNIPPET some-template-name }}

{{! END-SNIPPET }}
```

or in JS/TS

```typescript
// BEGIN-SNIPPET

// END-SNIPPET
```

It uses https://github.com/ef4/ember-code-snippet underneath, so anything mentioned in the docs should apply within e-c (though, we use 2.4.x, so we _do_ have syntax highlighting and do not have the `get-code-snippet` helper mentioned on the main page)

> * Does the "snippets" support `.ts` or `.d.ts`?

Yep, it should.

> * Should I be using decorators in the example or using classic syntax? I suppose the latter could work, but it doesn't work super well with TypeScript, but perhaps it works good enough for the purpose of the example.

Anecdotally, it seems most folks using TypeScript are using it with decorators, so we can probably stick to those.

> * Glimmer or classic component?

I'm assuming both are fairly similar? Glimmer would probably be okay on its own, but if there are notable differences, it might be good to note those.

> * For classic syntax at least, I believe `.get` is supposed to Just Works™ here, should I be using that?

Probably worth mentioning in the docs, but we should probably prefer using regular ES getters.

> I don't think it's particular dangerous, because at the end of the day it's just a weird quirk/fallout, but I'm not sure it's good to include it as an official API.
Got it. Sounds okay to leave out then.

chancancode:
I'm writing some docs, but sounds like @chriskrycho wants to review this also

maxfierke:
> I'm writing some docs, but sounds like @chriskrycho wants to review this also

awesome! his suggestion of type tests would be super valuable too, if that's a thing

addon/index.d.ts
```diff
@@ -0,0 +1,828 @@
+import ComputedProperty from '@ember/object/computed';
+
+export type TaskGenerator<T> = Generator<any, T, unknown>;
```
chancancode, line 3:
```suggestion
export type TaskGenerator<T> = Generator<any, T, any>;
```

Should the third parameter (the LHS of `yield ...`, i.e. what it "returns") be `unknown` or `any`?

Due to TS limitations, there is no way we can type that properly. What the `yield` "returns" is based on what you yielded, and that is just not possible to type. `await` has a similar problem but is hardcoded to work in the TS compiler.

The first parameter (the RHS of the `yield ...`, i.e. what you are allowed to yield to ember-concurrency) is typed to `any` because there is no real utility to limit it, it really can be anything (including nothing – you may just be yielding so ember-concurrency can pause to check for `isDestroying` and such).

The third parameter is typed to `unknown`, but I am not sure if that is too strict. In practice, it will force every consumer to check the value. If you are in the mood for it you may write an assertion:

```ts
class Foo {
  @task *myTask() {
    let foo /* unknown */ = yield someStringPromise;
    assert('foo should have resolved into a string!', typeof foo === 'string');
    // now foo is a string
  }
}
```

...that's fine when it's easy enough to check, but it's this is really checking for bugs in the library than in your code, and you would never have written that assertion there if not for TS. For more complex objects you may end up needing to write elaborate type assertion functions that are otherwise not needed. In practice, a lot of people is just going to end up writing this anyway, every single time they yield in a task:

```ts
class Foo {
  @task *myTask() {
    let foo: MyType = (yield someStringPromise) as any;
  }
}
```

If that's what everyone is going to do at the end of the day, it just seems a little silly not to type it as `any` in the first place. In practice, so long as you know the type of the RHS, there is very little reason not to trust e-c to call you with the correct thing on the LHS. If you are worried that you would have mistyped the LHS, you could do this:

```ts
type Resolved<T> = T extends PromiseLike<infer R> ? R : T;

class Foo {
  @task *myTask() {
    let foo: Resolved<typeof someStringPromise> = yield someStringPromise;
  }
}
```

That seems like a pretty good compromise between correctness and verbosity, and led me to thinking we should go with `any` instead of `unknown` here. Thoughts?

chancancode: Ah, I realized my mistake (I actually thought about it when I typed this), in practice it doesn't *really* matter what we typed here, because the types are never really used for inference in the position I was refereeing to above. Since it's inside the body of a generator function, TypeScript is just going to do whatever it normally does to generator functions without ever considering the type we have here. This type is only used when assigning a generator function to `task(...)`, etc, as long as the type is wide enough that TS allows assigning any valid task function to positions where we _take_ one as a `TaskFunction<T>`, it doesn't really matter.

addon/index.d.ts
chriskrycho: @chancancode can you file a types issue referencing this at https://github.com/typed-ember/ember-cli-typescript? Now that we're landing the bump in the core types to 3.7, and with some other work I have in mind to tackle in the next few months, we *should* be able to make `get` work correctly. (Conditional types!)

addon/index.d.ts
chriskrycho: I think I'm on the same page given the constraints TS gives us. I would recommend we *very* carefully document this, however, because it's the kind of thing that could easily bite people by letting `any` leak in when they're unaware of it.

chancancode: I was wrong _again_. Here is how it comes up in practice (typing the return type of the generator function) and the ergonomic differences:

[Playground](https://www.typescriptlang.org/play/index.html?experimentalDecorators=true&ssl=30&ssc=1&pln=31&pc=1#code/GYVwdgxgLglg9mABAQwM6oKYCcoAoIIAmMsCAXCmAJ4CUFamOqiBYxpSA3ogL4BQoSB0RQ0Aa1wA6aciwBzVPWoBtALp1KVRN358oVAA4ZEAZShYY0ACriA4hjDZkUOFgA8VgHyIAvInuOWM6ubsjUADSIVpHgYmBwAO5gngDceobGADJwcJg2qGIBTi7uXr7+DsUhYVSR0ZqpfOlGiABKGKhwADYAbhiEHt5+VogYAB5QDoTMAApYcAC2MJiZMGIYbjBgwNht3gD8bYgUVmnNxu1QIFhgW3KD5VIy8oqaajS+3qfnbR3dfYQrBkPKMJlNmJdrrcwPcap4hlFQZM2BCMFcbnc3HNFssMKt1pttrtWvDEIdWsdfuiwECjIMUogmhAugxEAAxHLaPiIHmIAACogKiAAVKhzJYoPkJBozBZrHZKkESm4xRYYd5ONzedqumjEAlEuVsUtMJIsH9ehhcAByA0Ja00NLanV682dS2EcpUGAYLqeu1O508hjYPD6IxwYCIN3-fq+Hx+a2qu4Os5Bnm6qCIMUgYBRvze32eqAAC2Wkhzedwjq1QZDOFw4YwkezVzz8cTyZhqaa6ejaKhrdzwEkLgAqgYjFgAMJoK0fADU-fdANHcAnU9nmGrgZ5um1AvEIq6OTy4mrFGyuQwUqKSpCXbkGtrzsz+sNfmNuLNFr6NrtPZ9m+MYehQ7Qrv0bhNi2doIoWfrvgkabpm+lbAGBv79LSGylssyhJm2wDWqocE+ghuGoBWhE7r26bmtSQ55muG7YFu86IEuIGruOk6sXONHavuvJoFQkCMcAF6IF+mAquK6pcn29GDgRw7WruvB8PwQA)

This doesn't look so bad since the types are just strings and the assertion is easy to write, but basically see my first comment for the tradeoffs.

chancancode: Talked to @chriskrycho on [Discord](https://discordapp.com/channels/480462759797063690/484421406659182603/717868405587312682), it seems like we are leaning towards using `any` here and documenting/recommending the `Resolved` type trick

chriskrycho:
As promised, [Typed Ember RFC: Type Stability for Addons](https://github.com/typed-ember/ember-cli-typescript/pull/1158).

addon/index.d.ts
chriskrycho: Notably, both the `Resolved` type trick *and* the `assert` approach will work if we type as `any` here: `assert` will correctly narrow both `unknown` and `any`.

chancancode: Right, the main question is do we want to _force_ the consumer to do a check.

chriskrycho: As you noted, it's a *very* frustrating experience to *have* to when the type *is* actually knowable and local. I don't have a strong opinion here (and I normally *very* strongly favor `unknown` and forcing the check).

cc. also @dfreeman @jamescdavis

chancancode: Maybe I'll just go ahead and write a compiler plugin that transforms `@task async` methods into generators and no have to worry about this 😄

chancancode: Anddd done: https://github.com/chancancode/ember-concurrency-async

chancancode:
Update: I have added the requested documentation, will work on the tests next. In the meantime please review what I wrote.

Rendered screenshot, for reference:

localhost_4201_docs_typescript

tests/dummy/snippets/ts/yield/ember-concurrency-async.d.ts
```diff
@@ -0,0 +1,25 @@
...
+type Return<T extends AnyAsyncTaskFunction> =
+  T extends (...args: any[]) => Promise<infer R> ? R : unknown;
+
+declare module 'ember-concurrency' {
+  function task<T extends AnyAsyncTaskFunction>(taskFn: T):
+    TaskProperty<Return<T>, Args<T>>;
+
+  function taskGroup<T extends AnyAsyncTaskFunction>(taskFn: T):
+    TaskGroup<Return<T>>;
+}
```
chancancode, line 20: Is there a good way for ember-concurrency-async to bundle this monkey-patch? I could include a `.d.ts` file, but I'm not sure if TypeScript will ever "finds" it, since most people have `node_modules` on "excludes" and you don't otherwise need to import anything else from the addon to use it.

tests/dummy/app/docs/typescript/template.hbs
```diff
@@ -0,0 +1,143 @@
+<h3>Using ember-concurrency with TypeScript</h3>
+
+<p>
+  As of version 1.2, ember-concurrency comes bundled with its own
```
chancancode, line 4: I assumed we are going to bump minor version for this.

tests/dummy/app/docs/typescript/template.hbs
```diff
@@ -0,0 +1,157 @@
...
+  <a href="https://github.com/machty/ember-concurrency/blob/master/addon/index.d.ts">
+  type definitions</a>. TypeScript should automatically pick them
+  up to provide type checking and completion.
+</p>
+
+<p>
+  Note that old versions of <a href="https://github.com/machty/ember-concurrency-decorators/">
+  ember-concurrency-decorators</a> came with an incomplete type
+  definition for ember-concurrency. If you are using that addon,
+  be sure to upgrade to the latest version to avoid conflicts.
```
chancancode, line 14: I assumed this as well. I'm not sure what the version number would be, so I just said "latest". There is a 1.1 "alpha" that is pretty unrelated, so we may want to skip 1.1 to avoid confusion. Jumping to 1.2 would also match up with the e-c version number, a happy accident.

addon/index.d.ts
chancancode: I went with `any` in the end.

tests/dummy/snippets/ts/yield/ember-concurrency-async.d.ts
chancancode: Still relevant cc @chriskrycho @dfreeman

addon/index.d.ts
```diff
@@ -0,0 +1,828 @@
...
+ *   doDishes:      task(taskFn).group('chores'),
+ *   changeDiapers: task(taskFn).group('chores')
+ * });
+ * ```
+ *
+ * @param taskFn The generator function backing the task.
+ * @returns {TaskGroup}
+ */
+export function taskGroup<T extends TaskFunction<any, any[]>>(taskFn: T):
+  TaskGroup<TaskFunctionReturnType<T>>;
```
chancancode, line 559: I am pretty sure this is wrong:

1. it does not (should not) take a `taskFn`
2. it returns a `TaskGroupProperty` instead of a `TaskGroup`

chancancode: https://github.com/machty/ember-concurrency/pull/358

chancancode:
Pushed a new commit to add the type tests, will work on the ember-try and travis config next.

tests/dummy/app/docs/typescript/template.hbs
chancancode: (still relevant)

addon/index.d.ts
chancancode: #358 is merged, and this has been rebased to add docs and tests

chancancode:
Rebased and updated to account for #358, I think this is good to go!

chriskrycho:
Thank you so much! I have this on my list for the week—it will likely be Tuesday or Wednesday before I have time to review it in detail!

tests/dummy/snippets/ts/yield/ember-concurrency-async.d.ts
dfreeman: The typical pattern is to include it in `index.d.ts` or whatever your `types` entrypoint is (or a module imported from there), and then instruct users to write e.g. `import 'ember-concurrency-async';` in a central place in their project.

chancancode: I can do that (and update this documentation) once this has been released

tests/dummy/snippets/ts/yield/ember-concurrency-async.d.ts
```diff
@@ -0,0 +1,30 @@
+import { Task, task, taskGroup } from 'ember-concurrency';
+import { taskFor } from './utils';
```
chancancode, line 2:
```suggestion
```

tests/dummy/snippets/ts/yield/ember-concurrency-async.d.ts
chancancode: https://github.com/chancancode/ember-concurrency-async/pull/2

addon/index.d.ts
```diff
@@ -0,0 +1,940 @@
...
+} | {
+  one(event: string, callback: (...args: any[]) => void): void;
+} | {
+  addEventListener(event: string, callback: (...args: any[]) => void): void;
+  removeEventListener(event: string, callback: (...args: any[]) => void): void;
+};
+
+type Resolved<T> = T extends PromiseLike<infer R> ? R : T;
+
+type Settlement<T> = { state: 'fulfilled', value: T } | { state: 'rejected', reason: any };
```
chriskrycho, line 596: We discussed the tradeoffs around `any` vs. `unknown` elsewhere, but on this one I'm more inclined toward `unknown`, as it "forces" the right response for users, in a good way. Thoughts?

addon/index.d.ts
```diff
@@ -0,0 +1,940 @@
...
+ *
+ * - if the task that `yield`ed `all()` is canceled, any of the
+ *   {@linkcode TaskInstance}s passed in to `all` will be canceled
+ * - if any of the {@linkcode TaskInstance}s (or regular promises) passed in reject (or
+ *   are canceled), all of the other unfinished `TaskInstance`s will
+ *   be automatically canceled.
+ *
+ * [Check out the "Awaiting Multiple Child Tasks example"](/docs/examples/joining-tasks)
+ */
+export function all<T extends readonly unknown[] | readonly [unknown]>(
```
chriskrycho, line 686: One note and one question here—

- If we want to support tuples (it's nice!), we should add support for some further (if limited) number of tuples, `[unknown, unknown]`, `[unknown, unknown, unknown]`, etc.—I'd suggest going out to four or five, as that's what we've generally done with Ember's own types.
- Why the `readonly`?

addon/index.d.ts
```diff
@@ -0,0 +1,940 @@
...
+ * The normal version of a `RSVP.allSettled` just returns a regular, uncancelable
+ * Promise. The `ember-concurrency` variant of `allSettled()` has the following
+ * additional behavior:
+ *
+ * - if the task that `yield`ed `allSettled()` is canceled, any of the
+ *   {@linkcode TaskInstance}s passed in to `allSettled` will be canceled
+ */
+export function allSettled<T extends readonly unknown[] | readonly [unknown]>(
+  values: T
+): Promise<{ -readonly [K in keyof T]: Settled<T[K]> }>;
```
chriskrycho, line 702: As above!

addon/index.d.ts
```diff
@@ -0,0 +1,940 @@
...
+ * additional behavior:
+ *
+ * - if the task that `yield`ed `race()` is canceled, any of the
+ *   {@linkcode TaskInstance}s passed in to `race` will be canceled
+ * - once any of the tasks/promises passed in complete (either success, failure,
+ *   or cancelation), any of the {@linkcode TaskInstance}s passed in will be canceled
+ *
+ * [Check out the "Awaiting Multiple Child Tasks example"](/docs/examples/joining-tasks)
+ */
+export function race<T>(values: readonly T[]): Promise<Resolved<T>>;
```
chriskrycho, line 754: Should this also have the extra tuple overloads?

addon/index.d.ts
```diff
@@ -0,0 +1,940 @@
...
+
+export interface EncapsulatedTaskDescriptor<T, Args extends any[]> {
+  perform(...args: Args): TaskGenerator<T>;
+}
+
+export type EncapsulatedTaskDescriptorArgs<T extends EncapsulatedTaskDescriptor<any, any[]>> =
+  T extends { perform(...args: infer A): TaskGenerator<any> } ? A : [];
+
+export type EncapsulatedTaskDescriptorReturnType<T extends EncapsulatedTaskDescriptor<any, any[]>> =
+  T extends { perform(...args: any[]): TaskGenerator<infer R> } ? R : unknown;
```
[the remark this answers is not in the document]
chancancode: Because the feature is called and documented as `EncapsulatedTask`. I considered just calling it `EncapsulatedTask `, but that's not quite right, because this is the thing you pass in (similar to the POJO you pass to `EmberObject.extend(...)`) to create a encapsulated task (technically, an encapsulated task property, which gets turned into an encapsulated task, which can be instantiated as encapsulated task instances). It just so happens that the current implementation of the `task` function normalizes away the "encapsulated-ness" of it so there aren't `EncapsulatedTaskProperty`, `EncapsulatedTask` and `EncapsulatedTaskInstance` exports, but there very well may be in the future.

As far as `TaskDescriptor`, that also sounds like too generic of a term, as it seems feasible that there will be a `TaskDescriptor` that is unrelated to `EncapsulatedTask`.

Ultimately, I question whether some of these APIs are still needed/still a good fit in the post-Octane world, but ultimately, it just doesn't seem like my decision to make or my job to speculate on the future direction too much in this PR, as my goal is just to document what exists today.

So `EncapsulatedTaskDescriptorArgs` seems like the most conservative, and accurate, if a bit verbose, option.

addon/index.d.ts
chancancode: I think I copied this from the built-in `Promise` type. Since `TaskInstance<T>` implements `Promise<T>`, I figured it's good to match the built-in promise things as close as possible to avoid any accidental divergances and assignability issues. The rest of it falls out from that decision.

addon/index.d.ts
chancancode: As far as I can tell, the `T1, T2, T3, ...` approach is obsolete. This is based on some of the newer built-in types in TS (I think this is based on the latest `Promise.all`).

There are tests for this, but basically the `unknown[]`  covers the other possibilities, while the `...[K in keyof T]...` in the return type will match everything nicely. I think `[unknown]` is needed as a special case otherwise it will get typed as `unknown[]` which is less precise and has some undesirable consequences. 

As far as `readonly`, admittedly it's mostly just because I copied it, but I believe that's because `readonly` is the narrowest possible thing here and it's what we actually require, so it's good to be precise, otherwise, if someone has an array that's marked readonly for some reasons, they won't be able to pass it in.

In general, I think it's a good idea to mark array-based APIs as `readonly` where possible for that reason.

addon/index.d.ts
chancancode: same

addon/index.d.ts
chancancode: same

addon/index.d.ts
chriskrycho: Interesting. I will need to dig in a bit and figure out what value the `| [unknown]` adds, because I can [confirm][playground] that it infers correctly with *just* `unknown[]`.

In any case, I’m happy enough to leave it given it’s copied from `Promise` directly. 👍 Thanks for clarifying!

[playground]: https://www.staging-typescript.org/play?#code/C4TwDgpgBAShDOB7ANgNwgEwDwBUB8UAvFDlBAB7AQB2G8UACgE6IC2AlvBADLsDWELO2oAzCE0YEA-IygAuEgG4AUMowQAxsgCGTaCICu1DcHaJqUbcmS4ylGnSh7tGc8hBQjfaogDu1AG0AXTwACmUoSKhUKwMEBRwAGmUASgVmNk5BAG8oAFpnV2p3KACAaShhKAEQRBESIIU4JDRMXHKQqABfPBVVdS1daGQIYEsFAPhgJmEAcyCVAZ09KBGxgCMFKZnqWeDFzWXh0agNCe25xKhqA1Z18QXVK2RQ7RTlZ9D198+NFKA

addon/index.d.ts
chriskrycho: Ah, very good; that makes sense. I dislike it, but that’s been there for a *long* time and would break a lot of existing code, so I understand why they’re leery of changing it. Thanks!

addon/index.d.ts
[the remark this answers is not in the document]
chriskrycho: This makes sense; thanks for the explanation! I had forgotten about the feature in question. 👍

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+  });
+
+  test('waitForProperty', async () => {
+    let obj = { foo: 'foo' };
+
+    // @ts-expect-error (it actually works – bug in expect-type?)
+    expect<typeof waitForProperty>().toBeCallableWith(obj, 'foo', 'bar');
+    waitForProperty(obj, 'foo', 'bar');
+
+    expect<typeof waitForProperty>().toBeCallableWith(obj, 'foo', v => v === 'bar');
```
chriskrycho, line 1559: `v` here needs a type annotation.

```suggestion
    expect<typeof waitForProperty>().toBeCallableWith(obj, 'foo', (v: unknown) => v === 'bar');
```

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+      let result = await waitForEvent(document.body, 'click');
+      expect<typeof result>().toEqualTypeOf<void>();
+    }
+  });
+
+  test('waitForProperty', async () => {
+    let obj = { foo: 'foo' };
+
+    // @ts-expect-error (it actually works – bug in expect-type?)
+    expect<typeof waitForProperty>().toBeCallableWith(obj, 'foo', 'bar');
```
chriskrycho, line 1556: 1. TS (and thus `expectTypeOf`) is legitimately catching an error in the type as written, because of a quirk in how TS handles resolution of overloads. If you place the `value` overload *last* in this list, this works correctly. I poked at it a bit, and what you need for the best set of resolutions and inference is: 

    ```ts
    export function waitForProperty<O extends object, K extends keyof O>(
      object: O, key: K, callbackOrValue: (value: O[K]) => boolean
    ): Yieldable;
    export function waitForProperty(
      object: object, key: string, callbackOrValue: (value: unknown) => boolean
    ): Yieldable;
    export function waitForProperty<O extends object, K extends keyof O>(
      object: O, key: K, callbackOrValue: O[K]
    ): Yieldable;
    ```
    
    If you do that, you'll be able to drop the `@ts-expect-error` on line 1561, and you'll also need to update line 1565 because the parameters resolved for the default overload will be different, but you'll no longer have these inference errors.

2. Marking this as an error to be ignored means that your *consumers* will have to do the same! Any time you hit a thing like this, feel free to ping one of the Typed Ember folks for input—ending up with something that consumers of the types have to work around as a cast is always 😬.

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+    });
+  });
+
+  test('classic ember (encpsulated task)', () => {
+    ClassicComponent.extend({
+      myTask: task({
+        foo: 'foo',
+
+        *perform(immediately: boolean, ms: number = 500) {
+          let self = this;
```
chriskrycho, line 1673: What does this binding of `this` to `self` accomplish in terms of the tests? 🤔 (No argument implied; at first glance I would *assume* it's identical, but I figure there's a reason you added it.)

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "noEmit": true,
+    "strict": true,
+    "baseUrl": ".",
+    "paths": {
+      "ember-concurrency": ["./addon/index.d.ts"]
+    }
+  },
+  "exclude":[
+    "node_modules",
+    "node_modules/ember-concurrency-decorators/types/ember-concurrency.d.ts",
```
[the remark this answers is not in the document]
chancancode: I am not 100% sure that it did anything, tbh. However, e-c-d ships with its own ambient (not imported by anything) declaration that conflicts with the definitions here. When importing from e-c-d, it seems to cause that file to be "picked up" and causes type errors in the language server in VS Code, which appears to be fixed by this, or that I got lucky after adding it.

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+    tp.readOnly();
+
+    // @ts-expect-error
+    tp.property();
+
+    // @ts-expect-error
+    tp.meta();
+
+    expect<typeof tp.on>().toBeCallableWith();
+    expect<typeof tp.on>().toBeCallableWith('init');
```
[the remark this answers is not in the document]
chancancode: no-op. I could type it as `(event: string, ...events: string[])` but the actual implementation is pretty happy to receive nothing: https://github.com/machty/ember-concurrency/blob/613dc1807f21d9a74886f9eb524cbc2b5e18ae86/addon/-task-property.js#L512-L513

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+    expect<typeof tp.on>().toBeCallableWith('init', 'didInsertElement');
```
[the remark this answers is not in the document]
chancancode: I don't think so, at least it's not documented and I have not personally used it that way: https://github.com/machty/ember-concurrency/blob/613dc1807f21d9a74886f9eb524cbc2b5e18ae86/addon/-task-property.js#L480-L510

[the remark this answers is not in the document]
chriskrycho: Very interesting. So this `on` is *not* like Ember Evented's `on`. 👍 Thanks for clarifying!

tests/types/ember-concurrency-test.ts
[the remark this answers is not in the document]
chriskrycho: Thanks for the implementation link. Quirky, but as long as the types match the implementation, 👍!

tests/types/ember-concurrency-test.ts
chancancode: Thanks for digging into this. I don't (or at least didn't) think the previous version was _actually_ a problem – that's why this line was followed by an actual call to the function to confirm that it works w/o type errors. Either way, it's good to have it fixed and thanks for the pointers.

tests/types/ember-concurrency-test.ts
chancancode: There _was_ a reason, but I don't remember what it was anymore. It has to do with `this` being a special type sometimes (hovering on it and the TSLS just shows `this: this`) but assigning it forces it to infer. But when I tired it again, it doesn't have the problem anymore so I fixed it. Maybe it was specific to TS3.6, or maybe the build will fail and I'll remember why. 🤷‍♂️

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+  });
+
+  test('Task', () => {
+    // @ts-expect-error
+    new Task(); // Task cannot be constructed
+
+    // @ts-expect-error
+    class Foo extends Task {} // Task cannot be subclassed
+
+    let t!: Task<string, [boolean, number?]>;
```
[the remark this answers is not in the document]
chancancode: This didn't work because `declare let` doesn't work in block scope

tests/types/ember-concurrency-test.ts
```diff
@@ -0,0 +1,1883 @@
...
+  rawTimeout,
+  task,
+  taskGroup,
+  timeout,
+  waitForEvent,
+  waitForProperty,
+  waitForQueue
+} from 'ember-concurrency';
+import { restartableTask } from 'ember-concurrency-decorators';
+import { expectTypeOf as expect } from 'expect-type';
```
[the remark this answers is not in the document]
chancancode: I did the search/replace but left the import name. I think for the most part, `expect(...).toBeString()` etc actually reads fine, and I think the main reason to emphasize on the `TypeOf` part is when used along side jest etc where `expect` is already taken as a name, since we aren't here, I think the shorter name is nice and makes it easier to read. (And I don't want to have to reindent a bunch of lines in case if width changes too much.)

tests/types/ember-concurrency-test.ts
chriskrycho: Ah, I see what you meant by the original comment now. You're correct that it was _callable_ the other way, and didn't show a type error, but when I inspected it directly it was resolving the wrong overload. 😬 Kind of weird, to be honest, but in this case I think `expectTypeOf` was catching something that likely would have shown up in a *really* weird way in the wild.

tsconfig.json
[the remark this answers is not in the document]
chancancode: In any case, I'll try to remember to come back and remove this once e-c-d has been updated

PR #362, opened 2020-06-19
Allow accessing encapsulated task state

chancancode:
From [discord](https://discordapp.com/channels/480462759797063690/484421754702397441/722746509363249192): 

> @andreyfel : Hi @chancancode! I'm so excited to try out the new typings for e-c! Could you please advise how to deal with encapsulated task and the inner properties? In this example http://ember-concurrency.com/docs/encapsulated-task/ you can access uploadFile.last.progress. If I try to do in from TS, the progress field doesn't exist in the uploadFile.last, because it is just a TaskInstance which doesn't have such field.

Had I known that this is a thing, I would probably have written the types a bit differently, but this should work and is non-breaking.

I didn't make `EncapsulatedTaskInstance` public because I had this lingering suspicious that we'll need to add a second generic parameter to `TaskInstance` for the args. I don't want to lockdown that possibility, or deal with strangely mismatched ordering of parameters between `TaskInstance` vs `EncapsulatedTaskInstance`, so I kept that private.

For the time being, the public way to type it (as seen in the tests) is `TaskInstance<T> & { ... }`, which I think is plenty acceptable.

Got to update e-c-async and e-c-ts to account for this, then update the Octane tests.

cc @andreyfel


addon/index.d.ts
```diff
@@ -332,16 +367,9 @@ export interface TaskInstance<T> extends Promise<T> {
...
- * A {@link TaskProperty} is the Computed Property-like object returned
- * from the {@linkcode task} function. You can call Task Modifier methods
- * on this object to configure the behavior of the {@link Task}.
- *
- * See [Managing Task Concurrency](/#/docs/task-concurrency) for an
- * overview of all the different task modifiers you can use and how
- * they impact automatic cancelation / enqueueing of task instances.
- */
-export interface TaskProperty<T, Args extends any[]> extends ComputedProperty<Task<T, Args>> {
+type EncapsulatedTaskInstance<T, State extends object> = TaskInstance<T> & EncapsulatedTaskState<State>;
```
chancancode, line 370: Note that this is private/unexported

package.json
```diff
@@ -43,9 +43,9 @@
     "ember-cli-template-lint": "^1.0.0-beta.1",
     "ember-cli-uglify": "^2.1.0",
     "ember-code-snippet": "^2.4.0",
-    "ember-concurrency-async": "^0.2.0",
+    "ember-concurrency-async": "chancancode/ember-concurrency-async#encapsulated-task-v2",
     "ember-concurrency-decorators": "^2.0.0",
-    "ember-concurrency-ts": "^0.1.0",
+    "ember-concurrency-ts": "chancancode/ember-concurrency-ts#encapsulated-task-v2",
```
chancancode, line 48: 🤷‍♂️ Not ideal, but these are development dependencies, and we have a chicken and egg problem. Either we merge it like this, once this is released as 1.2.1 I will release those and come back to update this. Or we merge this PR without the second commit, and I come back to update the Octane tests later.

tests/types/ember-concurrency-test.ts
```diff
@@ -714,58 +1217,76 @@ module('unit tests', () => {
     }
 
     {
-      let tp = task({ *perform() {} });
-      expect(tp).toEqualTypeOf<TaskProperty<void, []>>();
+      let tp = task({ foo: 'foo', *perform() {} });
+      expect(tp).toMatchTypeOf<TaskProperty<void, []>>();
```
chancancode, line 1221: The `toMatchTypeOf` here and in the rest of the file confirms that this is non-breaking, since it's still assignable.

tests/types/ember-concurrency-test.ts
```diff
@@ -88,12 +98,224 @@ module('unit tests', () => {
...
+    }
+  });
+
   test('Task', () => {
     // @ts-expect-error
     new Task(); // Task cannot be constructed
 
     // @ts-expect-error
-    class Foo extends Task {} // Task cannot be subclassed
+    class Foo extends Task<any, any[]> {} // Task cannot be subclassed
```
chriskrycho, line 318: This is my only concern with the tweaks here as I see it, and I *think* it’s fine because inference works correctly in the cases we *don’t* expect to be errors. 🤔 We could in principle set the type args to default with `= any` and `= any[]`, but that might have other undesirable knock-on effects? In general, that’s the move I tend to reach for in scenarios like this where you want to preserve backwards compatibility while needing to make this kind of change. However, again, I think it’s needless here?

tests/types/ember-concurrency-test.ts
```diff
@@ -486,6 +767,228 @@ module('unit tests', () => {
...
+
+    // @ts-expect-error
+    tp.cancelOn(false);
+
+    // @ts-expect-error
+    tp.cancelOn('init', false);
+
+    expect(tp.restartable).toBeCallableWith();
+    expect(tp.restartable).parameters.toEqualTypeOf<[]>();
+    expect(tp.restartable).returns.toEqualTypeOf<typeof tp>();
```
chriskrycho, line 825: Here and elsewhere where you’re writing it this way, can you not just write it like this?

```suggestion
    expect(tp.restartable).returns.toEqualTypeOf(tp);
```

addon/index.d.ts
```diff
@@ -20,16 +34,22 @@ export type EncapsulatedTaskDescriptorArgs<T extends EncapsulatedTaskDescriptor<
...
- * to create run individual {@linkcode TaskInstance}s,
- * and at any point, you can call the {@linkcode Task#cancelAll .cancelAll()}
- * method on this object to cancel all running or enqueued
- * {@linkcode TaskInstance}s.
- */
-export interface Task<T, Args extends any[]> {
+export type EncapsulatedTaskState<T extends object> = ExcludeKeys<
+  ExcludeKeys<T, TaskInstance<any>>,
+  { perform: any }
+>;
```
dfreeman, line 40: I think instead of building custom `ExcludeKeys` and `NonOverlappingKeys` types, this could be written as:

```ts
export type EncapsulatedTaskState<T extends object> = Omit<
  T,
  'perform' | keyof TaskInstance<any>
>;
```

andreyfel:
Hi @chancancode! Thank you for this PR! It seems that it is covering my use case! I've tested against that branch and it works!

The only thing I'm missing now is how to type `this` within the `perform` function of the encapsulated task. If I want to set a property on task instance inside perform what type does it have? EncapsulatedTaskInstance?

tests/types/ember-concurrency-test.ts
chancancode: This was not related to the PR. The genetics are already there and requires, this is trying to test that it errors because it cannot be subclasses, but was previously erroring because The generics are missing.

chriskrycho: Ah, very good—I thought it was odd that the changes you made would have required this. 😅

chancancode:
I don't think we should block on making `EncapsulatedTaskInstance` public. It is very rare that you would have to type it (please show some examples – most cases I know of are either implicit or can be inferred, and I think it's rare that you would want to use this in a parameter or return position).

I do think people will copy the type alias, and I think that's not ideal but not so bad. If we a public one, it will be compatible with and won't break these existing usages.

I do think we should work towards making it public, we just have to address the other blocker first–  whether we need to add the second generic to `TaskInstance` – and that's far from a hypothetical/theoretical issue. I don't think this PR is a good place to address that but we can also come to a decision on that fairy soon and fix both.

In the meantime, I think this adds enough values to those who need the feature and it's not obvious to me that they would be inconvenienced by this during normal usage.

chancancode:
> The only thing I'm missing now is how to type `this` within the `perform` function of the encapsulated task. If I want to set a property on task instance inside perform what type does it have? `EncapsulatedTaskInstance`?

Why did you have to type it? There are tests that confirms the `this` inference works correctly: https://github.com/chancancode/ember-concurrency/blob/encapsulated-task-v2/tests/types/ember-concurrency-test.ts#L2347-L2348

Is that not working for you for some reason?

andreyfel:
@chancancode, yes, it works in a basic case. But I have a little bit more complex case.
The idea is to have a task which exposes a property isRunningLong which should turn true in
case if the task takes longer than `loadingTimeout`. I'm using an encapsulated task with an inner task for it:

```
export interface LongRunningTaskDescriptor<T, Args extends any[]> extends EncapsulatedTaskDescriptor<T, Args> {
  isRunningLong: boolean,
  timeoutTask: TaskProperty<void, []>,
}

export function longRunningTask<T, Args extends any[]>(fn: TaskFunction<T, Args>, loadingTimeout = 300): LongRunningTaskDescriptor<T, Args> {
  return {
    isRunningLong: false,

    timeoutTask: task(function * (this: LongRunningTaskDescriptor<T, Args>) {
      yield timeout(loadingTimeout);
      set(this, 'isRunningLong', true);
    }).drop(),

    * perform(...args): TaskGenerator<T> {
      try {
        this.timeoutTask.perform();
        return yield * fn.apply(this.context, args);
      } finally {
        this.timeoutTask.cancelAll();
        set(this, 'isRunningLong', false);
      }
    },
  };
}
```

It is used like this with new e-c and e-c-d:
```
@task({ restartable: true })
fetchSmth = longRunningTask(function * () {
    yield fetchSmth();
}, 600)
```

So, I have to specify type of `this` for the `timeoutTask`. And inside the `perform` function this.timeoutTask.perform doesn't work. And `taskFor` doesn't work here.
Also typescript says that this.context doesn't exist. I'm not sure if it is a type issue or context is not a public filed.

maxfierke:
`context` is sort of "intimate", undocumented API. It's useful for implementing extensions onto e-c, so perhaps its worth making public but its not something we'd expect people to use often (though with encapsulated tasks, it might make more sense.)

I think a composition like that to create an encapsulated task is a bit of an advanced use-case, so if it's working in the basic case, it may be worth deferring for later unless addressing it is fairly straightforward.

PR #363, opened 2020-06-20
Types: Make Task, TaskInstance, and TaskGroup extend EmberObject

jamescdavis:
A `Task` is an `EmberObject`:
https://github.com/machty/ember-concurrency/blob/d066cd71a9e6abb8c749d931e971004a9930c587/addon/-task-property.js#L53-L55

This is important because the task state properties are computed properties, e.g.
https://github.com/machty/ember-concurrency/blob/d066cd71a9e6abb8c749d931e971004a9930c587/addon/-task-state-mixin.js#L14-L22
and require using `.get()` to access them in Ember < 3.1.

Update: `TaskInstance` and `TaskGroup` are also `EmberObject`s so I've included them as well.

@chancancode @dfreeman @chriskrycho

### DIFF
--- a/addon/index.d.ts
+++ b/addon/index.d.ts
@@ -235,7 +235,7 @@ export interface TaskGroup<T> {
  * because concurrency policy enforced by a
  * {@linkcode TaskProperty Task Modifier} canceled the task instance.
  */
-export interface TaskInstance<T> extends Promise<T> {
+export interface TaskInstance<T> extends Promise<T>, EmberObject {
   /**
    * If this TaskInstance runs to completion by returning a property
    * other than a rejecting promise, this property will be set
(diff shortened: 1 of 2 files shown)


jamescdavis:
AsyncTask types and tests for async tasks w/ task wrapper are here: https://github.com/jamescdavis/ember-concurrency-decorators/pull/1 but we obviously can't drop them in like that. Need to think about how to make them manually importable like `ember-concerrency-async`s are.

chancancode:
Should `TaskInstance` be updated too?

jamescdavis:
Yep!: https://github.com/machty/ember-concurrency/blob/d066cd71a9e6abb8c749d931e971004a9930c587/addon/-task-instance.js#L798

Also TaskGroup: https://github.com/machty/ember-concurrency/blob/2b9c8c40979ff42e2449fde0c9ca7e985e023691/addon/-task-group.js#L35

I can add these here.

PR #365, opened 2020-07-26
Implement `animationFrame()` yieldable

Turbo87:
This PR implements a `animationFrame()` utility that wraps the native [`requestAnimationFrame()` API](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) and can be used via `yield animationFrame()` (similar to `yield timeout(1000 / 60)`).


addon/index.d.ts
```diff
@@ -739,6 +739,32 @@ export function allSettled<T extends readonly unknown[] | readonly [unknown]>(
...
+ *
+ * The task below, when performed, will print the time since the last loop run
+ * for every animation frame.
+ *
+ * ```js
+ * export default Component.extend({
+ *   myTask: task(function * () {
+ *     let lastNow = performance.now();
+ *     while (true) {
+ *       yield animationFrame(1000);
```
maxfierke, line 754: we don't need this argument, right?

addon/utils.js
```diff
@@ -262,6 +262,52 @@ export function rawTimeout(ms) {
...
+ *
+ * The task below, when performed, will print the time since the last loop run
+ * for every animation frame.
+ *
+ * ```js
+ * export default Component.extend({
+ *   myTask: task(function * () {
+ *     let lastNow = performance.now();
+ *     while (true) {
+ *       yield animationFrame(1000);
```
maxfierke, line 295: same here?

addon/index.d.ts
Turbo87: whoops, you're totally right. copy-paste mistake 🤦

Turbo87:
> Could you add a quick test?

any suggestions on how to test this? 😅

maxfierke:
Could add a unit test that stubs out rAF (sinon should be available in the test suite) and asserts that the correct things are called and that the task runs to completion without error (might not even need to stub out rAF)

Adding to these existing tests too would be good:

* `tests/types/ember-concurrency-test.ts`, need a test for the types
* [`tests/unit/cancelable-promise-helpers-test.js#L392`](https://github.com/machty/ember-concurrency/blob/067fb79524b69ba4118765bb709ef93aec4800bc/tests/unit/cancelable-promise-helpers-test.js#L392) can add to that existing test to ensure the cancelation behavior works

Turbo87:
@maxfierke I've added some tests but I fail to understand why the TS typing tests are failing. They seem to work alright locally.

@chancancode any clues?

Turbo87:
@maxfierke @chancancode any ideas how to proceed? the failing tests seem unrelated to the code that I introduced here 🤔

maxfierke:
Looks like those typescript tests are also failing in https://github.com/machty/ember-concurrency/pull/368. I'll take a look at it this week and try and figure it out

maxfierke:
@Turbo87 master's passing now, if you want to give this a rebase?

Turbo87:
hmm, seems there is still an unrelated test failing. or maybe it's just flaky. can you restart the failed test to check?

maxfierke:
yeah, that's just a flaky one. I queued up a restart on that failed job but doesn't seem to be taking. I'll merge as-is

maxfierke (approved):

PR #368, opened 2020-08-25
Test ember-concurrency against embroider build

kiwiupover:


alexlafroscia:
I know this PR has been open/around a while, but Embroider now publishes a `test-setup` package that is meant specifically to abstract the set-up for testing Embroider away from addon developers. It would probably be better to use that instead!

https://github.com/embroider-build/embroider/tree/master/packages/test-setup

maxfierke:
embroider testing is now enabled and passing on master, so closing this out. Thanks!

PR #384, opened 2020-11-13
v2: Provide nice decorators via ember-concurrency-decorators

maxfierke:
This sets up some foundational work to begin combining [`ember-concurrency-decorators`](https://github.com/machty/ember-concurrency-decorators), by directly depending upon the current addon but without requiring separate import statements (or a separate dependency in a consuming app or addon's `package.json`)

It pulls in some base types for decoration options, and delegates the aliases for now to ember-concurrency-decorators. It also clones the decorator tests.

The "ugly" decorators (e.g. `(@(task(function* () { ... }).drop())` will remain available by virtue of coming "for free" with the computed-based TaskProperty implementation (necessary to support the classic EmberObject model for task hosts), but will be deprecated in favor these "nice" decorators.

A follow-up PR will convert all the docs to using the "nice" decorators.

A future PR (possibly post a final v2.0.0 release) will seek to fully integrate the "nice" decorators and remove the dependence on `computed` to provide decorator support for Ember versions that do not require it.

### DIFF
--- a/addon/index.d.ts
+++ b/addon/index.d.ts
@@ -1,3 +1,14 @@
+export {
+  restartableTask,
+  dropTask,
+  keepLatestTask,
+  enqueueTask,
+  restartableTaskGroup,
+  dropTaskGroup,
+  keepLatestTaskGroup,
+  enqueueTaskGroup
+} from 'ember-concurrency-decorators';
+
 import ComputedProperty from '@ember/object/computed';
 
 export type TaskGenerator<T> = Generator<any, T, any>;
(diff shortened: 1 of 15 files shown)


tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "target": "es6",
     "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
dfreeman, line 5: Was this needed to get things typechecking for some reason? Generally I don't think this flag's use is encouraged if you're not writing Angular code that relies on it.

addon/index.d.ts
```diff
@@ -648,6 +659,73 @@ type Settlement<T> = { state: 'fulfilled', value: T } | { state: 'rejected', rea
 
 type Settled<T> = Settlement<Resolved<T>>;
 
+// Decorator option types from ember-concurrency-decorators
```
dfreeman, line 662: I know the plan is eventually to bring e-c-d in, but in the meantime for consistency is there a reason not to re-export these types the same way you're re-exporting the runtime values?

maxfierke: I'm not sure, but they aren't exported in e-c-d, so I didn't want to expose them until I have an understanding for why they were not exported there.

tsconfig.json
maxfierke: Nope, that's leftover from trying something. Will remove.

maxfierke:
superseded by #386

PR #387, opened 2020-11-25
v2: Update docs to use native classes, decorators, Octane features, etc.

maxfierke:
Refactors the docs to use native classes, decorators, and other Octane features. It's not a full Octane/Glimmer Component-ification because there's certain things that aren't polyfillable back to Ember 3.8 LTS, but it gets close.

Notably, the tutorial pages (Writing Code Without Tasks, Post-Mortem, Refactoring With Tasks) have had their examples refactor to show re-writing of `async`/`await` code to Tasks instead of `Promise` code (`async`/`await` not being as widespread when these examples were written in 2016/2017.)


esbanarango:
This is great!! 🔝

tests/dummy/app/components/task-function-syntax-5/component.js
```diff
@@ -1,30 +1,24 @@
 // BEGIN-SNIPPET task-function-syntax-5
-import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
+import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { task } from 'ember-concurrency';
 
-export default class MyOctaneComponent extends Component {
-  @tracked status = null
+export default Component.extend({
```
machty, line 6: What's happening here? Seems like a revert to old school Ember?

maxfierke: Yeah, instead of documenting Octane use (which we do throughout), I wanted to maintain a documented use with Classic Ember.

PR #407, opened 2021-02-25
Add missing jsdoc documentation

maxfierke:
In moving stuff around, some methods such as `cancelAll` got lost in the jsdoc documentation. Went through and made sure all public API was adequately documented in jsdoc.

h/t @Turbo87 for the heads up on the `cancelAll` and @bendemboski who noticed `onState` hadn't been documented

### DIFF
--- a/addon/-private/task-decorators.js
+++ b/addon/-private/task-decorators.js
@@ -150,7 +150,7 @@ export const lastValue = decoratorWithParams((target, key, descriptor, [taskName
  * method.
  *
  * ```js
- * import Component from '@glimmer/component';
+ * import Component from '@ember/component';
  * import { task } from 'ember-concurrency';
  *
  * class MyComponent extends Component {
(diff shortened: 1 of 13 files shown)


addon/-private/task-decorators.js
```diff
@@ -128,37 +128,286 @@ export const lastValue = decoratorWithParams((target, key, descriptor, [taskName
...
+ *
+ * ```js
+ * import Component from '@glimmer/component';
+ * import { task, dropTask } from 'ember-concurrency';
+ *
+ * class MyComponent extends Component {
+ *   &#64;task
+ *   *plainTask() {}
+ *
+ *   &#64;dropTask({ cancelOn: 'click' })
```
Turbo87, line 203: does this work? if this is a Glimmer component then it wouldn't have a `click()` hook, right?

maxfierke: Oh yeah good catch, I'll move this back to an ember component.

PR #408, opened 2021-03-02
Issues using async/await with stubbed promises

BarryThePenguin:
👋🏻 

This is my first expedition into the `ember-concurrency` source 👀 
`ember-concurrency` is a joy to use, especially with the new TypeScript additions

I've put together a reproduction of an issue I ran into while upgrading to ember 3.24

There's two tests..

1. "sinon stub promise that fails"
2. "sinon stub promise that passes"

In both an async method is stubbed with `sinon.stub(obj, 'sinonPromise').rejects('An error')`.

Then the test calls `await obj.tryDoStuff()` which performs a task that yields the stubbed method.

In the first test I can see `asyncErrorsHandled` is not set on the task.

That's about as much information as I can provide 😅

### DIFF
--- a/tests/unit/task-test.js
+++ b/tests/unit/task-test.js
@@ -9,6 +9,7 @@ import Ember from 'ember';
 import { task, timeout, forever } from 'ember-concurrency';
 import { module, test } from 'qunit';
 import { destroy } from '@ember/destroyable';
+import sinon from 'sinon'
 import { decoratorTest } from '../helpers/helpers';
 
 const originalLog = console.log;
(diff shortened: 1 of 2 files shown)


BarryThePenguin:
I think this might be closely related to #409 too..

maxfierke:
@BarryThePenguin is there a way you can reduce this down to a case without sinon? If there's a generalizable reproduction, it might make it easier to determine what's going on without having to disentangle what sinon might be doing (as much as I like sinon, there's fair amount of indirection, typically)

I agree, it does sound like it might be similar to #409 but a generalizable test case can help us confirm

BarryThePenguin:
I don't know if the latest changes provide much more info.. but it's essentially doing the same as `sinon.stub()`, but we're replacing functions on the instance manually instead..

maxfierke:
There seems to be a variance based on version, indicating perhaps there's some dependency difference at play. I ran the tests against master:

- It passes on Ember 3.8, 3.16, 3.20, and 3.22
- It fails on Ember 3.12, release (currently 3.25), beta (currently 3.26), canary, and... 3.22 (when run through ember-try, instead of directly)

AFAICT, there haven't been any updates to RSVP or Backburner that I would typically expect would impact this, but I'll keep digging

PR #413, opened 2021-03-08
RFC: Public Yieldables API

maxfierke:
[Rendered](https://github.com/machty/ember-concurrency/blob/mf-public_yieldable/rfcs/0000-yieldables.md)

Implementation is included as well, which shows refactor of built-in Yieldables to use the new API. Refactoring (or similar) will probably stand, regardless of whether to promote to public API, as calling `taskInstance.proceed` with various magic values in a bunch of places is error prone.

### DIFF
--- a/rfcs/0000-yieldables.md
+++ b/rfcs/0000-yieldables.md
@@ -74,16 +74,16 @@ class YieldableState {
   return(value: any): void;
 
   // Raise a given error within the given task instance and halt execution
-  throw(error: Error): void;
+  throw(error: any): void;
 }
 
-abstract class Yieldable {
+abstract class Yieldable<T> implements PromiseLike<T> {
   constructor(...args) {
     // User setup logic would go here. If the yieldable took arguments, they
     // could be stored in the Yieldable here
   }
 
-  onYield(state: YieldableState): Function<void> {
+  onYield(state: YieldableState): () => void {
     // This is where most user code would go, defining what happens when the
     // task encounters `yield myYieldable`.
 
--- a/rfcs/0000-yieldables.md
+++ b/rfcs/0000-yieldables.md
@@ -1,0 +1,179 @@
+---
+Stage: Accepted
+Start Date: 2021-03-07
+Release Date: Unreleased
+Release Versions:
+  ember-concurrency: vX.Y.Z
+RFC PR: 
+---
+
+<!--- 
+Directions for above: 
+
+Stage: Leave as is
+Start Date: Fill in with today's date, YYYY-MM-DD
+Release Date: Leave as is
+Release Versions: Leave as is
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
+
+# Yieldables
+
+## Summary
+
+Promote Yieldables to public API, as a way to instrument TaskInstances by
+providing a safe mechanism to implement custom waiters, hooks, introspection,
+and other operations in a task definition, from application code.
+
+## Motivation
+
+> Why are we doing this? What use cases does it support? What is the expected
+> outcome?
+
+There are many great uses for yieldables, and indeed many of them are already
+built-in to ember-concurrency today. Yieldables like `timeout`, `animationFrame`,
+and `rawTimeout` provide safe, cancelation-aware ways to delay execution of a
+task. Cancelable Promise helpers like `all`, and `hashSettled` are too a form of
+yieldable.
+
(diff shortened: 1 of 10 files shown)


rfcs/0000-yieldables.md
```diff
@@ -0,0 +1,179 @@
...
+For example, if I wanted to implement a yieldable for `requestIdleCallback`, I
+could do the following:
+
+```javascript
+import Component from '@glimmer/component';
+import { task, Yieldable } from 'ember-concurrency';
+
+class IdleCallbackYieldable extends Yieldable {
+  onYield() {
+    this.callbackId = requestIdleCallback(() => this.continue());
```
Turbo87, line 98:
```suggestion
    this.callbackId = requestIdleCallback(() => this.next());
```

or am I misreading the other snippet above?

maxfierke: Whoops, yeah, I was going based on implementation and not proposal 🤦  I like `.next` better for parity with generator functions, so will fix that and update the implementation.

machty:
Looks great! Glad you're making this more public. I like that some of the ugly internals for "resuming" within the yieldable are hidden.

One thing though:

> In general, Yieldable instances should not be reused across calls

I would change this so that you _can_ resume a yieldable instance across calls. A Yieldable should be more like a stateless "description" of the async operation you want to build when e-c processes the Yieldable you yield to it. This is the basic idea behind Rx Observables: the observable objects don't have any state on them until you `.subscribe()` to it, and it allows for patterns of composability that are pretty neat (i.e. transform/combine a bunch of observables into one single observable that you subscribe to). I'm not sure those same composability benefits would come to the Yieldables API in the same way as it does for observables, but it seems like a better foundation to build on then one where yieldables can't be reused between yields.

From what I gather, I don't think making this change would actually change the public API for _building_ yieldables, and it has the nice benefit that you _would_ be able to reuse yieldable instances across calls, e.g.

```
const sharedTimeout = timeout(500);

// ...

yield sharedTimeout;
// do stuff
yield sharedTimeout;
```

vs having to wrap `sharedTimeout` in a function.

maxfierke:
> Looks great! Glad you're making this more public. I like that some of the ugly internals for "resuming" within the yieldable are hidden.
> 
> One thing though:
> 
> > In general, Yieldable instances should not be reused across calls
> 
> I would change this so that you _can_ resume a yieldable instance across calls. [...]
> 
> From what I gather, I don't think making this change would actually change the public API for _building_ yieldables, and it has the nice benefit that you _would_ be able to reuse yieldable instances across calls, e.g.
> 
> ```
> const sharedTimeout = timeout(500);
> 
> // ...
> 
> yield sharedTimeout;
> // do stuff
> yield sharedTimeout;
> ```
> 
> vs having to wrap `sharedTimeout` in a function.

@machty I think I might have phrased this weirdly... it is safe to use across calls within the same TaskInstance execution, but not across different TaskInstances. i.e. yes, you can make one call to `timeout` and yield that multiple times within the same task instance, but you could not, say, do the following:

```javascript
const myTimeout = timeout(500);

class MyClass { 
  @task *aTask() {
    yield myTimeout
  }
  
  @task *bTask() {
     yield myTimeout;
   }
}
```

^ This would not be allowed.

If we want to allow that, I think we can do that, although the API might need to be a little bit different, especially to accomidate Promise casting of yieldables (which I just realized I forgot to define.)

MLeganes:
Hallo there,

I am getting a lots of emails from you guys. This is because I was ussing your packed in my app when I was working in my company and now I am now working anymore and I would like to offer myself to help you. I have some experience with js and ember.js but I will need to learn a bit about concurrency. Will you need my help? Just let me know if it possible to help.

Best regards,
MLeganes

maxfierke:
> I am getting a lots of emails from you guys. 

@MLeganes it looks like you're "watching" the repo and thus are probably getting notifications for everything. If this is not what you want, you can unsubscribe by clicking "unwatch" at the top of the page or changing your notifications in that dropdown, or see this GitHub support article: https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/configuring-notifications

maxfierke:
@machty Revised the RFC last night a little bit in order to introduce `YieldableState` to keep the `TaskInstance` invocation state separate from the `Yieldable` instance's state. Would be curious if you have any additional thoughts?

Otherwise I think I'll probably move forward with this approach.

PR #414, opened 2021-03-16
Remove runloop binding for perform helper

### LINKED ISSUE #409: Async Error Handler Running Synchronously
I'm running into a pretty bizarre error while trying to upgrade our app to `ember-concurrency@2.0.2` (from the latest pre-`2.0.0` release).

It seems that the async error handler defined here is in fact running synchronously, and prempting our own error handling.

We have a `swallow-error` helper that we sometimes use for cases where an Ember Concurrency task would expect an error to be thrown and we use the derived error state on the task instance to show it to the user. Because we error will be "handled" in the UI itself, we want to swallow the error entirely.

<details>
  <summary><code>swallow-error</code> helper source</summary>
  <p>

```js
import { helper } from '@ember/component/helper';

/**
 * This helper is meant to be wrapped around another action that might throw an error that we want to suppress.
 *
 *   ```hbs
 *   <button {{on 'click' (swallow-error (perform this.someTaskThatMightThrow))}}>
 *     Click Me
 *   </button>
 *   ```
 */
export function swallowError([fn]) {
  return function callAndSwallowError(...args) {
    try {
      const response = fn(...args);

      if (response.catch) {
        return response.catch(function () {
          // Swallow async error
        });
      }

      return response;
    } catch (e) {
      // Swallow synchronous error
    }
  };
}

export default helper(swallowError);
```

</p></details>

In the current version of `ember-concurrency` that we're using, this works just fine.

In `ember-concurrency@2.0.2`, it seems like the code here that should report the error if it was not already caught is being executed synchronously when it _should_ be executed asynchronously

https://github.com/machty/ember-concurrency/blob/35a188cd86d322ae0b0c443d5b0c5cb009b1793c/addon/-private/external/task-instance/executor.js#L344-L348

I know it's a challenge to grok a stack trace without much context, but this is what I'm seeing; a single synchronous trace from my `debugger` point inside the error-reporting callback that _should_ be run asynchronously straight down to the function invocation inside `callAndSwallowError`.

CleanShot 2021-03-03 at 16 35 40@2x

I know that `env.async` is using `join` from the `@ember/runloop` package, which will insert the callback into the current queue. My guess is that Backburner is calling things in the current queue before proceeding with execution of the code in my helper... for some reason. My code isn't doing anything directly with the runloop; there's just the normal amount of wrapping that helper invocations have by default at play.

I tried to create a failing test case in the `ember-concurrency` repository but have been unable to get the same error to occur there just yet. I will post a PR with that once I have it.

I'm wondering whether a potential fix could be specifically scheduling the `async` callback into the _next_ run loop, rather than using `join` which can insert it into the _current_ "tick".

maxfierke:
Wrapping `perform` in a full runloop means that a synchronous task throw will be
reported within that same invocation, before another helper is able to wrap and
instrument the task instance. Since it's already run, it's too late to do things
like attaching catch handlers. We do not need this runloop binding anymore, since
there's no auto-run assertion with the versions supported by ember-concurrency
2.x.

Fixes #409

### DIFF
--- a/addon/-private/external/task-instance/executor.js
+++ b/addon/-private/external/task-instance/executor.js
@@ -6,13 +6,13 @@ import {
   YIELDABLE_THROW,
   YIELDABLE_RETURN,
   YIELDABLE_CANCEL,
-  cancelableSymbol
-} from '../yieldables';
+  cancelableSymbol,
+} from "../yieldables";
 
 import {
   COMPLETION_SUCCESS,
   COMPLETION_ERROR,
-  COMPLETION_CANCEL
+  COMPLETION_CANCEL,
 } from "./completion-states";
 import {
   CancelRequest,
@@ -20,7 +20,7 @@ import {
   CANCEL_KIND_LIFESPAN_END,
   CANCEL_KIND_PARENT_CANCEL,
   didCancel,
-  TASK_CANCELATION_NAME
+  TASK_CANCELATION_NAME,
 } from "./cancelation";
 
 export const PERFORM_TYPE_DEFAULT = "PERFORM_TYPE_DEFAULT";
@@ -191,10 +191,10 @@ export class TaskInstanceExecutor {
   handleYieldedUnknownThenable(thenable) {
     let resumeIndex = this.index;
     thenable.then(
-      value => {
+      (value) => {
         this.proceedChecked(resumeIndex, YIELDABLE_CONTINUE, value);
       },
-      error => {
+      (error) => {
         this.proceedChecked(resumeIndex, YIELDABLE_THROW, error);
       }
     );
@@ -270,7 +270,7 @@ export class TaskInstanceExecutor {
       return;
     }
     this.disposers = [];
-    disposers.forEach(disposer => disposer());
+    disposers.forEach((disposer) => disposer());
   }
 
   /**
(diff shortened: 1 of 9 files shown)


maxfierke:
@alexlafroscia mind giving this a spin against your app to see if it resolves the problem "in the wild", so to speak?

alexlafroscia:
Can do! I'll check this branch out and see what happens

addon/-private/external/task-instance/executor.js
```diff
@@ -6,21 +6,21 @@ import {
   YIELDABLE_THROW,
   YIELDABLE_RETURN,
   YIELDABLE_CANCEL,
-  cancelableSymbol
-} from '../yieldables';
+  cancelableSymbol,
+} from "../yieldables";
```
alexlafroscia, line 10: I noticed this too when working in the repo -- it might be nice to add Prettier to the linter config. It's harder working in repos where it _isn't_ present than to work in ones where it is!

alexlafroscia (approved):
This solved the problem I was having in my app 🎉

addon/-private/external/task-instance/executor.js
maxfierke: Definitely! Big fan of prettier, I just hate setting it up myself 😆

PR #415, opened 2021-03-17
chore: add Prettier to JS lint config

alexlafroscia:
Making good on my offer here

https://github.com/machty/ember-concurrency/pull/414#discussion_r596158086

This adds Prettier to run as part of ESLint, and then "fixes" all of the files.

This does _not_ introduce Prettier for non-JS files (like Markdown or YAML) but that's something I could also add, if you're interested!

This also does not add a lint-on-commit hook, which I could set up as well if you're interested.

I went and added the default configuration here, which uses double-quotes rather than single-quotes. If anyone has a preference for single quotes instead, I can tweak that and update this PR. That would definitely change less files, since _most_ (but not all) used single quotes 😅


tests/dummy/app/components/tutorial-2/component.js
```diff
@@ -1,26 +1,28 @@
...
 export default class Tutorial2 extends TutorialComponent {
   result = null;
   isFindingStores = false;
 
   @action
   async findStores() {
-    if (this.isFindingStores) { return; } // ++
+    if (this.isFindingStores) {
+      return;
+    } // ++
```
maxfierke, line 14: We should make sure to add the `// ++` annotations to the other two lines above

tests/dummy/app/components/tutorial-3/component.js
```diff
@@ -1,28 +1,32 @@
...
+    this.set("isFindingStores", true);
 
-    let coords = await geolocation.getCoords()
+    let coords = await geolocation.getCoords();
     let result = await store.getNearbyStores(coords);
 
-    if (this.isDestroyed) { return; } // ++
+    if (this.isDestroyed) {
+      return;
+    } // ++
```
maxfierke, line 26: Same thing here!

tests/dummy/app/components/tutorial-4/component.js
```diff
@@ -1,33 +1,40 @@
...
 
     let geolocation = this.geolocation;
     let store = this.store;
 
-    this.set('isFindingStores', true);
+    this.set("isFindingStores", true);
 
-    try {                                   // ++
-      let coords = await geolocation.getCoords()
+    try {
```
maxfierke, line 21: We should make sure the `// ++` lines in this try/catch block line up here too to the lines that are added in this step of the tutorial

maxfierke (commented):
This looks really great @alexlafroscia! Just flagged a couple spots where it affected the meaning of some of the tutorial snippets.

alexlafroscia:
Thanks for the review @maxfierke! I missed the notification earlier, but I'll fix those up and push again(

alexlafroscia:
Hmmm... So, Prettier doesn't like the placement of those comments and pushes them onto other lines...

Maybe we have Prettier ignore the tutorial components?

maxfierke:
> Maybe we have Prettier ignore the tutorial components?

yeah, that'd be okay with me!

alexlafroscia:
Sorry for the delay in getting back to this!

I started over again since the code has changed some. Those `tutorial-*` components have been ignored so their formatting shouldn't have changed this time!

I went and configured Prettier to use single-quotes rather than double. It seemed like _more_ files did it that way than using double quotes, though there's a mix of both in here so that was pretty arbitrary. If you'd rather we use the default setting instead -- double quotes -- that would be easy to fix!

maxfierke (approved):
:octocat: looks great! Will spot-check the docs output this evening just to be safe, but otherwise I have no concerns!

PR #434, opened 2021-07-29
Expose TaskInstance state inside encapsulated tasks

maxfierke:
While TaskInstance state was exposed externally via proxying, the proxy was not used internally making TaskInstance state impossible to access. Fixes an issue reported in Discord.


jkeen:
What incredible timing, I just stumbled onto this bug while using `2.1.0` and here it is fixed. Thanks!

PR #438, opened 2021-08-27
Extract generic TaskFactory base class

maxfierke:
Some internal refactoring to separate the Ember bits of `TaskFactory`. This will provide some groundwork for a future RFC.


maxfierke:
CI fixes will be sizeable, so merging for now and will post seperate PR to get master back to green

Issue #463, opened 2022-05-23
Error: Accessing __ec_cancel__ is deprecated.


Turbo87:
Ember Data lately started throwing errors like this:

```
Error: Accessing __ec_cancel__ is deprecated.  Only available methods to access on a promise returned from model.save() are .then, .catch and .finally
    at handleDeprecationWorkflow (http://localhost:7357/assets/vendor.js:77665:15)
    at HANDLERS.<computed> (http://localhost:7357/assets/vendor.js:44403:9)
    at deprecationCollector (http://localhost:7357/assets/vendor.js:77702:5)
    at HANDLERS.<computed> (http://localhost:7357/assets/vendor.js:44403:9)
    at http://localhost:7357/assets/test-support.js:7547:10
    at HANDLERS.<computed> (http://localhost:7357/assets/vendor.js:44403:9)
    at invoke (http://localhost:7357/assets/vendor.js:44415:9)
    at Object.deprecate (http://localhost:7357/assets/vendor.js:44371:28)
    at Object.get (http://localhost:7357/assets/vendor.js:94927:18)
    at TaskInstanceExecutor.handleYieldedValue (http://localhost:7357/assets/vendor.js:151531:36)
```

It seems that the combination of using ember-concurrency with recent ember-data releases is the problem.

PR #464, opened 2022-05-23
Ignore "Accessing __ec_cancel__ is deprecated" errors from Ember Data

Turbo87:
Resolves https://github.com/machty/ember-concurrency/issues/463

/cc @maxfierke

### DIFF
--- a/addon/-private/external/task-instance/executor.js
+++ b/addon/-private/external/task-instance/executor.js
@@ -233,9 +233,9 @@ export class TaskInstanceExecutor {
       return;
     }
 
-    this.addDisposer(yieldedValue[cancelableSymbol]);
+    this.addDisposer(getSymbol(yieldedValue, cancelableSymbol));
 
-    if (yieldedValue[yieldableSymbol]) {
+    if (getSymbol(yieldedValue, yieldableSymbol)) {
       this.invokeYieldable(yieldedValue);
     } else if (typeof yieldedValue.then === 'function') {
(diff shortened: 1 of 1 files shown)


maxfierke (commented):
@Turbo87 does replacing this with a `obj.hasOwnProperty(symbol)` check work? That might be preferable to swallowing an exception

Turbo87:
```
Error: Accessing hasOwnProperty is deprecated.  Only available methods to access on a promise returned from model.save() are .then, .catch and .finally
```

unfortunately, that does not seem to work either :-/

/cc @runspired in case you have any suggestions

maxfierke:
> ```
> Error: Accessing hasOwnProperty is deprecated.  Only available methods to access on a promise returned from model.save() are .then, .catch and .finally
> ```
> 
> unfortunately, that does not seem to work either :-/
> 
> /cc @runspired in case you have any suggestions

I would argue this is breaking an assumption that you should be able to call methods that a regular `Promise` would still have (which `hasOwnProperty` would be one, since it inherits it from `Object`)

However, if this is in a shipped/release version of Ember Data, we should move forward with the `try/catch`

Turbo87:
I just realized that this is causing issues for us because we are using https://github.com/mixonic/ember-cli-deprecation-workflow with `throwOnUnhandled: true`, which turns regular deprecation warnings into `throw theDeprecation`. Ignoring the deprecation instead fixes the issue too, so there is actually no need for changing anything in ember-concurrency :)

PR #465, opened 2022-05-26
task() + async arrow fns

machty:
Implements Part 4 of [this RFC](https://github.com/machty/ember-concurrency/issues/46https://github.com/machty/ember-concurrency/issues/462), specifically, a new form of `task()` that accepts an async arrow fn that gets babel-transpiled into a generator function. This addresses a number of TS issues surrounding decorators and generator functions.


addon/-private/external/task-decorators.js
```diff
@@ -138,35 +138,3 @@ export const lastValue = decoratorWithParams(
     };
   }
 );
-
```
machty, line 141: NOTE: I deleted these because these are actually declared [in this file](https://github.com/machty/ember-concurrency/blob/364be8818cf98975cf878adaf9b430e676acc60f/addon/-private/task-decorators.js) -- these decorator declarations are completely unused.

PR #477, opened 2022-08-24
Simplify async fn API to just `task(async () => {})`

machty:
Having just released the new API for `task(this, async () => {})`, it dawned on me that with a bit more babel transform magic, we can support an even simpler task API of just:

```
myTask = task(async () => {
  // ...
})
```

rather than having to pass in `this` as the first arg. This seems like an obvious additional win.

### DIFF
--- a/addon/-private/async-arrow-runtime.js
+++ b/addon/-private/async-arrow-runtime.js
@@ -1,18 +1,12 @@
 import { TaskFactory } from './task-factory';
 
 /**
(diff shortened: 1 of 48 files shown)


jamescdavis:
Can you deploy the docs changes? 🙂

machty:
@jamescdavis just pushed it now, thanks for the reminder. I can't believe how long the short-lived `this`-based documentation was live :(

PR #479, opened 2022-08-25
Fix waitForProperty types, uncomment type tests

machty:
As part of the work for version 2.3 I forgot that when I bumped some dependencies, I started getting some type failures in `tests/types/ember-concurrency-test.ts`, so I commented those tests out.

This PR attempts to

1. re-enable those specs 
2. Fix the `waitForProperty` type to allow for the 3rd arg to be optional


machty:
I'm not sure what dependency might have changed that caused things to break, but here's the package.json diff: https://gist.github.com/machty/8394d88d2f9f29630ec0dd1e8c48fd6b

image

machty:
I'm not sure what dependency might have changed that caused things to break, but here's the package.json diff: https://gist.github.com/machty/8394d88d2f9f29630ec0dd1e8c48fd6b

image

PR #487, opened 2022-09-02
Fix observers in 2.3.0+

### LINKED ISSUE #481: [Bug] 2.3.1 broke `observes` modifier
### 🐞 Bug description

With the latest version of `ember-concurrency` the _observes_ modifier stopped working.

```js
import Model, { attr } from '@ember-data/model';
import { inject as service } from '@ember/service';
import { task } from 'ember-concurrency';

export default class Parking extends Model {
  @service store;

  @attr('array', { defaultValue: () => [] }) carIds;

  // This tasks should perform everytime `cardIds` changes.
  @task({
    observes: 'carIds.[]',
    drop: true,
  })
  *loadCarsTask() {
    // ...
  }
}
```

```js
import Component from '@glimmer/component';
import { action, set } from '@ember/object';

export default class ParkingBox extends Component {
  @action
  onCarChange(cars) {
    const { parking } = this.args;
    const cardIds = cars.map(({ id }) => parseInt(id));

    set(parking, 'cardIds', cardIds);
  }
}

```
### 😕 Actual Behavior
`loadCarsTask` isn't performing when `carIds` changes.

### 🤔 Expected Behavior
`loadCarsTask` should perform everytime `cardIds` changes.

### 🌍 Environment
- ember-concurrency  `2.3.1`
- ember-source  `4.6.0`
- ember-data `4.2.0`

machty:
Closes #481

It seems #457 removed a call to `_setupEmberKVO`, which was causing observers (and probably some other cases) to stop working.

### DIFF
--- a/addon/-private/external/task-decorators.js
+++ b/addon/-private/external/task-decorators.js
@@ -1,7 +1,7 @@
 import { TaskFactory } from './task-factory';
 
 function taskFromPropertyDescriptor(
-  _target,
+  target,
   key,
   descriptor,
   params = [],
@@ -23,6 +23,7 @@ function taskFromPropertyDescriptor(
   let tasks = new WeakMap();
   let options = params[0] || {};
   let factory = new factoryClass(key, taskFn, options);
+  factory._setupEmberKVO(target);
 
   return {
     get() {
--- a/tests/unit/decorators-test.js
+++ b/tests/unit/decorators-test.js
@@ -1,5 +1,6 @@
 import { module } from 'qunit';
 import { run } from '@ember/runloop';
+import EmberObject from '@ember/object';
 import { setOwner } from '@ember/application';
 import {
   task,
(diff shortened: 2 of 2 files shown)


addon/-private/external/task-decorators.js
```diff
@@ -23,6 +23,7 @@ function taskFromPropertyDescriptor(
   let tasks = new WeakMap();
   let options = params[0] || {};
   let factory = new factoryClass(key, taskFn, options);
+  factory._setupEmberKVO(target);
```
maxfierke, line 26: I think the idea was to keep anything ember-specific out of the `external` module, but if we're not doing the separate Ember descriptors, this seems reasonable for now, especially if we're planning on getting rid of observer support. We can also figure out a refactor later if we need to.

maxfierke (approved):
:octocat: lgtm

addon/-private/external/task-decorators.js
machty: 👍  Yeah I think we can revisit later

esbanarango:
Thank you @machty and @maxfierke 🙏

PR #504, opened 2022-11-28
Update example with a generic type on a task

ombr:
Hi,

I am not really sure how we should make this work with typescript but at least here is a failing test.

Luc


machty:
Doing some housekeeping; if you want to revive this, could you please add a breaking test to https://github.com/machty/ember-concurrency/blob/master/packages/test-app/types-tests/ember-concurrency-test.ts ?

Issue #510, opened 2023-01-09
Ember 4.8 types preview issue


machty:
https://blog.emberjs.com/announcing-official-typescript-types-public-preview/

```
../node_modules/ember-concurrency/addon/index.d.ts:379:11 - error TS2315: Type 'ComputedProperty' is not generic.

379   extends ComputedProperty<T> {
              ~~~~~~~~~~~~~~~~~~~

../node_modules/ember-concurrency/addon/index.d.ts:558:47 - error TS2315: Type 'ComputedProperty' is not generic.

558 export interface TaskGroupProperty<T> extends ComputedProperty<TaskGroup<T>> {
```

Issue is due to ComputedProperty's Get type parameter being removed. We don't have a solution yet but see here for further discussion:

https://discord.com/channels/480462759797063690/484421406659182603/1062000162341396481

gitKrystan:
I was tinkering with this today and came to the same conclusion as you did in https://discord.com/channels/480462759797063690/484421406659182603/1062000162341396481 : Trying to fix this within e-c's types breaks `get`'s "unwrapping" of task properties into tasks.

Since the native types dropped support for the old object model, which I believe will break the unwrapping anyway, I'm wondering if there is any remaining benefit to inheriting from `ComputedProperty` anymore. AFAICT this is only done for `get` anyway: https://github.com/machty/ember-concurrency/blob/624d747f82db20691ab70724dabe89233f367601/tests/types/ember-concurrency-test.ts#L797-L798

PR #512, opened 2023-02-07
Use the types published from Ember itself

gitKrystan:
One way to resolve https://github.com/machty/ember-concurrency/issues/510

This breaks the "unwrapping" of Tasks when using `get`.

Since we are no longer getting the benefits of using `ComputedProperty` for unwrapping of `get`, tomorrow I will add another commit removing that hack.


tests/types/ember-concurrency-test.ts
```diff
@@ -925,99 +925,99 @@ module('unit tests', () => {
     let O = EmberObject.extend({
       tp,
 
-      foo() {
-        let t = this.get('tp');
+      foo(this: EmberObject) {
+        let t = this.get('tp') as Task<string, [boolean, number?]>;
```
gitKrystan, line 929: With preview types, `get` generally returns `unknown`. I used casts to get around that. It's not pretty.

gitKrystan:
Relevant: https://github.com/emberjs/ember.js/pull/20387#issuecomment-1450074350

ef4:
I'm using this patch in an app and it's working well.

tests/types/ember-concurrency-test.ts
```diff
@@ -3214,6 +3242,106 @@ module('integration tests', () => {
...
+          return 'wow';
+        }
+      );
+
+      async performMyTask() {
+        // FIXME: According to @dfreeman this should work
+        // typeof myTask is `this["myTask"]`
+        let myTask = get(this, 'myTask');
+
+        expect(myTask).not.toBeAny();
```
gitKrystan, line 3313: This has the following type error:

```
Argument of type '[]' is not assignable to parameter of type 'MismatchArgs<IsAny<this["myTask"]>, false>'.ts(2345)
```

gitKrystan:
@dfreeman made a comment on a related PR (https://github.com/emberjs/ember.js/pull/20387#issuecomment-1450074350) that I looked into today.

> ...the type of `this.foo` and `get(this, 'foo')` should be the same once you take `ComputedProperty` out of the equation, so if `this.someTask.perform(...)` works, then `get(this, 'someTask').perform(...)` should also work. That's what [this signature](https://github.com/gitKrystan/ember.js/blob/7d3fb15b7d23d206ab2ca1b304c30a3f2496c75b/types/preview/%40ember/object/index.d.ts#L43) should be accomplishing.

Most of the type tests in this library were written pre-Octane, so they use `EmberObject.extend({})` and `task(function* () {})` which returns the type `TaskProperty`. In this case, while `this.foo` and `this.get('foo')` _do_ have the same type as indicated in Dan's comment, the type is `TaskProperty` rather than `Task`. This is because the `task` function returns a "computed-like" `TaskProperty` when passed a generator. In some cases, this can be handled with the `taskFor` function from [`ember-concurrency-ts`](https://github.com/chancancode/ember-concurrency-ts). In other cases, I needed to use casts.

In the newer Octane tests that use the async arrow function syntax, `this.foo` has the correct type. `get(this, 'foo')` is still not working properly, and I don't understand why. The returned type seems to be `this['foo']` which TS seems to be translating to `any`??? (Or at least, the `not.toBeAny()` is failing with a cryptic message.) See [here](https://github.com/machty/ember-concurrency/pull/512/files#diff-4e32e10f72cb7965e226fc102bc669d9224172741635e7f8ead7da659ccf433fR3308-R3314) (The test for "async arrow with get").

dfreeman:
> the type is TaskProperty rather than Task

I think the key piece that I was trying to communicate in that comment that may have gotten lost in the noise is this:

> instead of `TaskProperty extends ComputedProperty<Task<...>>`, I think `ember-concurrency` should be able to directly do `TaskProperty extends Task<...>`

In other words, anywhere in _any_ old code that's been using `ComputedProperty<T>`, we essentially now want to just use `T`, for all the reasons we talked about over in that PR about `get` no longer doing anything special. Do note that all of this discussion only applies to pre-Octane `Object.extend(...)` stuff (or at least conceptually it does; depending on how the specific types are entangled with one another in `ember-concurrency` there could be some cross-edition impact)

> The returned type seems to be `this['foo']` which TS seems to be translating to any??? (Or at least, the `not.toBeAny()` is failing with a cryptic message.) 

I think this is the intersection of two facts:
 - TS handles `this` field access in quirky ways with utilities like `get` due to the potential for those fields to be overridden in child classes with narrower types
 - It's really _really_ hard within the type system to discern whether something is `any` or not, since the whole point of `any` is to be a total chameleon—this means the `toBeAny` check can behave strangely in odd cases like this.

If actually invoking `perform` on the result of the `get` call works, then I think you're good 👍 

> `ms: number | undefined = 500`

This is suuuuper weird that it's inferring `ms: any` without the annotation. I'm guessing it's to do with how inference is working with all the `AsyncTaskArrowFunction` stuff, but I have no clue how/why that would have changed when swapping out the DT types for the native ones, since it doesn't seem like any of this inference actually depends on the Ember types.

gitKrystan:
@dfreeman I pushed some changes addressing your comment.

I believe the issue with defaulted arguments becoming `any` predates these changes because I was able to reproduce it on my project that uses the latest e-c release.

I've been investigating how to fix it, but no luck so far.

PR #515, opened 2023-03-23
Ember 4 Upgrade

elwayman02:
* Bump to latest github action versions
* Add Ember 4.4 scenario, drop scenarios below 3.28
* Switch from npm-run-all to concurrently
* Add ember-source peerDep
* Add node 18 support
* ember-modifier v3 (v4 is available but may require additional migration work)
* Added ember-sinon-qunit to make sinon play nice with ember-auto-import
* Bump volta to latest node/yarn minor/patch versions
* Convert assert.equal to assert.strictEqual or assert.deepEqual
* Autofix additional lint errors and disable 2 rules with weird errors
* Update caniuselist db

```
1..197
# tests 197
# pass  196
# skip  1
# todo  0
# fail  0

# ok
```


.template-lintrc.js
```diff
@@ -2,4 +2,8 @@
 
 module.exports = {
   extends: 'recommended',
+  rules: {
+    'no-whitespace-for-layout': false,
+    'require-input-label': false,
```
elwayman02, line 7: Turned off these two new rules because they had a couple violations that didn't make sense to fix:

1) A test rendered an input without a label
2) 2 of the docs templates threw "extra whitespace" errors but it seemed like a bug. I couldn't fix it.

elwayman02: Only 3 violations exist between the two of them, so they could be reenabled pretty easily if someone wants to attempt a fix.

elwayman02:
Actually, I believe ember-qunit and some other dependencies are finally forcing a drop of support for <3.28. I'll go ahead and remove these scenarios as it was going to be a major version bump anyway, and it's probably about time to drop support for Ember versions that have long been EOL'd.

SergeAstapov:
> We can probably drop node 10/12 support as well as part of this major release. What do you think?

@elwayman02 we should do it in a separate PR, right?

elwayman02:
Yea, it doesn't have to be done in this PR as long as the CI builds pass
once someone enables them.

elwayman02:
@machty are you able to take a look at this? It would be great to get tests actively running against an explicit 4.x LTS scenario soon. :)

Issue #521, opened 2023-04-06
Docs code examples gone in the latest version


esbanarango:
Looks like the latest version (v3(?)) broke the docs. All code examples are gone:

<img width="1043" alt="ember-concurrency" src="https://user-images.githubusercontent.com/1156865/230474720-505cef28-d4b1-4aa1-8c97-2938850eac0f.png">

PR #522, opened 2023-04-11
Fix code blocks not rendering on documentation site

camerondubas:
Code blocks on https://ember-concurrency.com/docs/tutorial are not rendering. This is due to deprecated code in the `CodeSnippet` component's template (provided by [ember-code-snippet](https://github.com/ef4/ember-code-snippet)).

To solve this, I bumped to thes latest `ember-code-snippet`, which required adding and configuring [ember-prism](https://github.com/shipshapecode/ember-prism) as well as creating a local `CodeSnippet` component as this is no longer provided by `ember-code-snippet` directly.

Additionally, the calculation of the max height for the template-toggle component broke so I refactored to use ResizeObserver.

Here are a couple of updated screenshots.  Note that the styling is a bit different from before. This can be customized, however, I felt this could be addressed in a follow-up PR.
image
image

Closes https://github.com/machty/ember-concurrency/issues/521


tests/dummy/app/components/code-snippet/component.js
```diff
@@ -0,0 +1,20 @@
...
+export default class CodeTemplateToggleComponent extends Component {
+  tagName = '';
+
+  @computed('name')
+  get snippet() {
+    return getCodeSnippet(this.name);
+  }
+
+  get language() {
+    if (this.snippet.language === 'handlebars') {
```
camerondubas, line 14: ember-prism expects hbs files to use the language "markup".  This feels a bit hacky. Would love a better solution

tests/dummy/app/components/code-template-toggle/component.js
```diff
@@ -10,40 +9,45 @@ export default class CodeTemplateToggleComponent extends Component {
...
-        })
-      );
-
-      let toggle = element.querySelector('.code-template-toggle');
+    const element = document.getElementById(this.id);
+    if (!element) {
+      return;
+    }
 
+    let resizeObserver = this._createResizeObserver((maxHeight) => {
```
maxfierke, line 21: do we need to call `disconnect()` on the observer in `willDestroyElement`?

maxfierke (commented):
Looks great! One question about observer cleanup.

(Looks like the embroider jobs are failing since #515 was merged, so ignore for now.)

tests/dummy/app/components/code-template-toggle/component.js
camerondubas: Good point! Updated it with the cleanup in `willDestroyElement`

PR #524, opened 2023-04-12
Improve styling of code blocks on documentation site

camerondubas:
Follow up to #522.

Cleaning up the styling of code blocks now that ember-prism is used.

Before: 
image

After:
 
image

### DIFF
--- a/tests/dummy/app/styles/app.scss
+++ b/tests/dummy/app/styles/app.scss
@@ -201,6 +201,16 @@ th, td {
   color: #bbb;
 }
 
+pre > code {
+  border: none;
+  margin: 0;
+  padding: 0 .5rem;
+}
+
+:not(pre) > code[class*="language-"], pre[class*="language-"] {
+  background-color: #f8f8f8;
+}
+
 .code-template-toggle {
   position: relative;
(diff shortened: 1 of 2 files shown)


tests/dummy/app/styles/app.scss
```diff
@@ -201,6 +201,16 @@ th, td {
...
+pre > code {
+  border: none;
+  margin: 0;
+  padding: 0 .5rem;
+}
+
+:not(pre) > code[class*="language-"], pre[class*="language-"] {
+  background-color: #f8f8f8;
+}
+
```
maxfierke, line 213: Is there more constrained scoping we could do here, maybe a class applied to the `CodeBlock` component?

camerondubas: Updated to use a class for styling the `code` element, but the `CodeBlock` doesn't provide a way to add a class to the `pre` directly. Instead, I updated the background color of the `.code-template-toggle` to match the prism theme, rather than updating the theme's background color.

The "after" screenshot in the PR description shows how this looks.

maxfierke (approved):

PR #526, opened 2023-04-27
Fix this is undefined in benchmark.js

ombr:
Benchmark task-modifier was not working out of the box on my tasks. I had to make sure we are using the right this. I was inspired by what is done in ember-concurrency-retryable:

https://github.com/maxfierke/ember-concurrency-retryable/blob/4b18ccc073872d8a7592c4f1eb6c17461938334b/addon/-private/retryable-task-instance.js#L72


machty:
I am interested in merging this... if you have time could you "rebase" against master? You'd have to move the code changes to the packages/test-app folder.

Issue #513, opened 2023-02-14
Typescript question for creating custom task modifiers


LanceStasinski:
Hello, I've added a custom task modifier to restore the functionality of `waitFor` from `@ember/test-waiters`. Do you have any suggestions for improving the types for this custom modifier? Thank you in advance!

```ts
import { TaskFactory, registerModifier } from 'ember-concurrency'
import { waitFor } from '@ember/test-waiters'

declare module 'ember-concurrency' {
  interface TaskFactory<T, Args> {
    taskDefinition: Function & TaskDefinition<T, Args>
  }
}

function waitForTask<T, Args extends unknown[]>(taskFactory: TaskFactory<T, Args>, option: boolean) {
  if (!option) {
    return
  }

  const { taskDefinition } = taskFactory
  const waitForTaskDefinition = waitFor(function* (this: unknown, ...args: Args) {
    yield* taskDefinition.apply(this, args)
  })

  taskFactory.setTaskDefinition(waitForTaskDefinition)
}

export function initialize() {
  registerModifier('waitFor', waitForTask)
}

export default {
  initialize,
}

```

SergeAstapov:
that will be great to bring this functionality back.
I wonder if it's worth to built into `ember-concurrency` itself

NullVoxPopuli:
> I wonder if it's worth to built into ember-concurrency itself

it should, yeah. `@ember/test-waiters` will not be integrating with ecosystem libraries (this expands scope too greatly)

PR #531, opened 2023-06-17
Add `waitFor` support to async arrow function API

bendemboski:
I am opening this PR to start a discussion. There is currently no way I can find to wrap async arrow function tasks in a test waiter a la `waitFor` from `@ember/test-waiters`. So this is a stab at adding support for that -- see the commit message for an explanation of this solution.

I haven't updated any documentation or anything because I want to see if this is a direction we want to pursue first.

A couple of alternatives I thought of:
1. Figure out how to make the babel plugin smarter so it can recognize when the task function is wrapped in a "modifier function," e.g. `myTask = task(waitFor(async () => { /* ... */ }));` would produce `{ generator: waitFor(function*() { /* ... */ }) }`. This seems sketchy because in the source code the modifier function would be called with an `async` function while in the transpiled code it would be called with a generator function. This happens to work for `waitFor`, but providing generic support for modifier functions like this would come with a pretty screwy caveat that they must support being passed generator functions even though they must be typed to accept `async` functions.
1. Make the options interface more generic -- define a type/interface for these sorts of modifier functions like `waitFor` and then do something like `myTask = task({ modifiers: [ waitFor ] }, async () => { /* ... */ });`. This would allow us to type the modifier functions to accept a generator function without running into trouble because the un-transpiled source code would never _invoke_  the modifier function. On the other hand, it complicates the syntax and it's not clear there will ever be a need for other modifier functions like this. I'm happy to explore this route if we prefer -- I don't think it would be much more work.
1. Just say we no longer support wrapping tasks in test waiters this way, and tell users to either wrap each promise in a `waitForPromise()` call, or work with the `@ember/test-waiters` maintainers to try to devise some other easy way to wrap `ember-concurrency` tasks.

I'm very interested to hear folks' thoughts as my app relies heavily on `@waitFor`, and I really love the async arrow function syntax and want to move to it soon as I can figure out what to do about the `@waitFor`s!

Related to #513

### DIFF
--- a/addon/-private/async-arrow-runtime.js
+++ b/addon/-private/async-arrow-runtime.js
@@ -16,6 +16,13 @@ export function buildTask(contextFn, options, taskName, bufferPolicyName) {
 
   const result = contextFn();
 
+  if (optionsWithBufferPolicy && optionsWithBufferPolicy.waitFor) {
+    result.generator = optionsWithBufferPolicy.waitFor(result.generator);
+
+    optionsWithBufferPolicy = Object.assign({}, optionsWithBufferPolicy);
+    delete optionsWithBufferPolicy.waitFor;
+  }
+
   const taskFactory = new TaskFactory(
     taskName || '<unknown>',
     result.generator,
--- a/addon/index.d.ts
+++ b/addon/index.d.ts
@@ -796,6 +796,16 @@ type OptionTypeFor<T, F> = F extends (...args: infer Args) => T
 type TaskOptions = OptionsFor<TaskProperty<unknown, unknown[]>>;
 type TaskGroupOptions = OptionsFor<TaskGroupProperty<unknown>>;
 
+type AsyncArrowFunctionTaskOptions<
+  HostObject,
+  T,
+  Args extends any[]
+> = TaskOptions & {
+  waitFor?: (
+    fn: AsyncArrowTaskFunction<HostObject, T, Args>
+  ) => AsyncArrowTaskFunction<HostObject, T, Args>;
+};
+
 type MethodOrPropertyDecoratorWithParams<Params extends unknown[]> =
   MethodDecorator &
     PropertyDecorator &
@@ -919,7 +929,7 @@ export function task<
 
 export function task<
   HostObject,
-  O extends TaskOptions,
+  O extends AsyncArrowFunctionTaskOptions<HostObject, any, any[]>,
   T extends AsyncArrowTaskFunction<HostObject, any, any[]>
 >(
   hostObject: HostObject,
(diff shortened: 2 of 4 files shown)


tests/types/ember-concurrency-test.ts
```diff
@@ -3128,6 +3129,7 @@ module('integration tests', () => {
       debug = task(this, { debug: true }, async () => {});
       onState = task(this, { onState: () => {} }, async () => {});
       onStateNull = task(this, { onState: null }, async () => {});
+      waitFor = task({ waitFor }, async () => {});
```
maxfierke, line 3132: I think this is missing the `this` (otherwise, it's the same test case as the one further down):

```suggestion
      waitFor = task(this, { waitFor }, async () => {});
```

addon/-private/async-arrow-runtime.js
```diff
@@ -16,6 +16,13 @@ export function buildTask(contextFn, options, taskName, bufferPolicyName) {
 
   const result = contextFn();
 
+  if (optionsWithBufferPolicy && optionsWithBufferPolicy.waitFor) {
+    result.generator = optionsWithBufferPolicy.waitFor(result.generator);
+
+    optionsWithBufferPolicy = Object.assign({}, optionsWithBufferPolicy);
+    delete optionsWithBufferPolicy.waitFor;
+  }
+
```
maxfierke, line 25: I think this probably should be moved into a modifier (_a la_ https://github.com/machty/ember-concurrency/issues/513), so that it works regardless of whether a task is running through the AyncArrowRuntime or not

addon/index.d.ts
```diff
@@ -796,6 +796,16 @@ type OptionTypeFor<T, F> = F extends (...args: infer Args) => T
...
+type AsyncArrowFunctionTaskOptions<
+  HostObject,
+  T,
+  Args extends any[]
+> = TaskOptions & {
+  waitFor?: (
+    fn: AsyncArrowTaskFunction<HostObject, T, Args>
+  ) => AsyncArrowTaskFunction<HostObject, T, Args>;
+};
+
```
maxfierke, line 808: This might not be needed to be special-cased if moved to a `waitFor` modifier

maxfierke (commented):
Overall, I like the approach. I think it's probably a reasonable expansion of the API as there's potentially other implementations someone might want to provide in advanced cases, so I appreciate it not being strictly coupled to the `@ember/test-waiters` implementation

I do think it should be implemented as modifier like the one suggested in #513, with the exception of its `option` being the `waitFor` function to use instead of a boolean.

NullVoxPopuli:
> or work with the @ember/test-waiters maintainers to try to devise some other easy way to wrap ember-concurrency tasks.

It's not `@ember/test-waiters`'s responsibility to integrate with ecosystem libraries -- that's backwards! 🙈 

allowing a task to be wrapped in _something_ is good though. May not even have to be a `waitFor`. could be "measureDuration" or some other utilities

acorncom:
Looks like the work discussed here was ultimately handled in https://github.com/machty/ember-concurrency/pull/536, can we close this PR?

bendemboski:
Yes, agreed

PR #537, opened 2023-07-22
Fix CI build

bendemboski:
Something (probably a release of new embroider packages) caused most of the floating dependencies scenarios to start failing in all branches. It looks like it was some kind of mismatch between the version of `@embroider/macros` used by `@ember/render-modifiers` (a dependency of `ember-prism`) and whatever version of `@embroider/shared-internals` it was finding in the `node_modules` tree.

So this commit updates us to the latest `@embroider/macros`, which allows us to dedupe `@embroider/macros` and `@embroider/shared-internals`, which appears to fix the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@ember/string": "^3.0.1",
     "@ember/test-helpers": "^2.9.3",
     "@ember/test-waiters": "^3.0.0",
-    "@embroider/macros": "^0.48.1",
+    "@embroider/macros": "^1.13.0",
     "@embroider/test-setup": "^2.1.1",
     "@glimmer/component": "^1.1.2",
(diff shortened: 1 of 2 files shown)


bendemboski:
The `embroider-safe` and `embroider-optimized` failures were [happening in master](https://github.com/machty/ember-concurrency/actions/runs/4738657764) before the failures this PR addressed started happening, so I believe this PR makes sense to merge as-is because it addresses one (newer) issue even though it doesn't address the other (older/pre-existing) issue.

maxfierke (approved):
:octocat: lgtm! I'll merge this as long as it's just the `embroider-*` ones failing (those were failing before, so don't feel compelled to fix those)

PR #538, opened 2023-07-22
Fix embroider scenarios

bendemboski:
Work around https://github.com/embroider-build/embroider/issues/1557 by moving the types tests out of `tests`


bendemboski:
FYI, I did double-check that the types tests are still running in their new location

maxfierke (approved):
:octocat: Ahhhh nice! I banged my head against the wall a couple times with this one

PR #539, opened 2023-08-01
Update compatibility indicators to Ember 4+

acorncom:
Two commits for this only because I’m using GitHub’s UI on my phone and it isn’t easy to amend a commit 😉

there may be other places this needs to be updated, these are the ones I’ve noticed so far


acorncom:
Looks like I may have misread the CI matrix here, don't remember much context on this. Closing for now

PR #551, opened 2024-02-03
Convert to V2 Addon

machty:
Notes:

- Minimum version is 3.28 now after getting rid of macros that were used to preserve pre-tracked computed property infrastructure


.github/workflows/ci.yml
```diff
@@ -51,7 +52,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        browser: [chrome, firefox]
+        browser: [chrome]
```
NullVoxPopuli, line 55: Something go wrong with Firefox testing?

packages/test-app/ember-cli-build.js
```diff
@@ -34,19 +34,26 @@ module.exports = function (defaults) {
       // GitHub Pages uses this filename to serve 404s
       emptyFile: '404.html',
     },
+
+    babel: {
+      plugins: [
+        require.resolve(
+          'ember-concurrency/babel-plugin-transform-ember-concurrency-async-tasks.js'
```
NullVoxPopuli, line 41: That's a long plugin name!

Thoughts on?:

"ember-concurrency/babel"?

It'd require a a prefix to use without require.resolve, like:

module:ember-concurrency/babel

packages/ember-concurrency/babel.config.json
```diff
@@ -0,0 +1,12 @@
+// Babel config for building the ember-concurrency package itself. Because EC is a V2 addon,
+// this babel config should *not* apply presets or compile away ES modules.
+// It exists only to provide development niceties for you, like automatic
+// template colocation.
+{
+  "plugins": [
+    // https://github.com/embroider-build/embroider/issues/1014
+    ["@babel/plugin-proposal-decorators", { "legacy": true }],
+    "@babel/plugin-proposal-class-properties",
```
NullVoxPopuli, line 9: Any reason not to use the current babel config from the blueprint?

.github/workflows/ci.yml
machty: Firefox builds are erroring with

> Error: no DISPLAY environment variable specified

Haven't figured out why, will try and remedy before merge.

https://github.com/machty/ember-concurrency/actions/runs/7770826285/job/21191457704

packages/test-app/ember-cli-build.js
machty: Ah yeah, I meant to rename this. Will DM you about `module:...`, don't know too much about that.

packages/ember-concurrency/babel.config.json
machty: I updated to the latest blueprint, thank you!

packages/test-app/ember-cli-build.js
machty: I shortened it; seems to work fine with `require.resolve`:

```
    babel: {
      plugins: [
        require.resolve('ember-concurrency/async-arrow-task-transform'),
      ],
    },
```

.github/workflows/ci.yml
machty: Restored ff testing; core issue was that I forgot there was a testem.js file that launched FF in headless mode that I forgot to move to test-app folder.

PR #569, opened 2024-02-22
Update assert in task-public-api.js to include note about required Babel transform for arrow functions

johanrd:


LucasHillDex:
Good change, this got me too.

machty:
Released in 4.0.1, thank you!

Issue #570, opened 2024-03-01
Having @glint/template as a non optional peer dependency causing type issues


LucasHillDex:
We are not currently using glint in our ember apps, and it seems most other packages in the ember ecosystem have `@glint/template` as an optional peer dependency. Upgrading to ember-concurrency 4.0 is causing type issues because of complex glint type incompatibilities that seem to stem from us not having our components fully registered in the glint registry. Is this intentional or should this be an optional peer dependency?

PR #572, opened 2024-03-05
Make glint template an optional peer dependency

LucasHillDex:
See issue: https://github.com/machty/ember-concurrency/issues/570

All other ember projects that have @glint/template as a peer dependency have it as optional as far as I can find. We are not using glint, and this imports types that cause issues in our codebase.


LucasHillDex:
@machty CI didn't kick off but I think I fixed it, thanks!

machty:
Thank you!

machty:
Released in 4.0.2

PR #573, opened 2024-03-19
Remove re-exports for helpers

bertdeblock:
Probably an oversight during the v1 -> v2 conversion?

### DIFF
--- a/packages/ember-concurrency/app/helpers/cancel-all.js
+++ b/packages/ember-concurrency/app/helpers/cancel-all.js
@@ -1,1 +1,0 @@
-export { default } from 'ember-concurrency/helpers/cancel-all';
--- a/packages/ember-concurrency/app/helpers/perform.js
+++ b/packages/ember-concurrency/app/helpers/perform.js
@@ -1,1 +1,0 @@
-export { default } from 'ember-concurrency/helpers/perform';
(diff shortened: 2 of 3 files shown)


machty:
Hmm how come we don't want this? Don't we still want ember-cli apps to have `{{perform ...}}` `{{cancel-all ...}}` and `{{task ...}}` to be available in pre-`.gts` components?

bertdeblock:
That's what these are for => https://github.com/machty/ember-concurrency/blob/master/packages/ember-concurrency/package.json#L121-L124

bertdeblock:
The entries in the `package.json` file are auto-generated AFAIK, based on https://github.com/machty/ember-concurrency/blob/master/packages/ember-concurrency/rollup.config.mjs#L23.

This is done via a custom Rollup plugin in `@embroider/addon-dev`.
Code that does the mapping => https://github.com/embroider-build/embroider/blob/main/packages/addon-dev/src/rollup-app-reexports.ts#L33.

The `_app_` folder is generated by Embroider, based on the defined app reexports and is the v2-addon equivalent of the `app` folder in v1 addons.

You can see it in ember-concurrency's `dist` folder => https://www.npmjs.com/package/ember-concurrency?activeTab=code

PR #574, opened 2024-04-02
Update README.md to include v2 addons

aklkv:


TSenter:
It'd be great to see this merged, I just encountered the same problem myself and wasn't entirely sure how to fix it.

PR #578, opened 2024-09-04
Fix missing @babel/helper-module-imports dependency

simonihmig:
This came up as that dependency imported from the transform [here](https://github.com/machty/ember-concurrency/blob/master/packages/ember-concurrency/async-arrow-task-transform.js#L19) not getting resolved in a strict pnpm environment (no hoisting).

### DIFF
--- a/packages/ember-concurrency/package.json
+++ b/packages/ember-concurrency/package.json
@@ -101,6 +101,7 @@
     "test": "tests"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


simonihmig:
Hey @machty, could you have a look at this please? It just came up again for us...

machty:
Thanks, released/published 4.0.3 with this dependency.

PR #610, opened 2026-06-20
Support Babel 8 in async-arrow-task-transform

johanrd:
## Problem

`async-arrow-task-transform` calls `api.assertVersion(7)`, which throws under `@babel/core` 8 (`Requires Babel "^7.0.0-0", but was loaded with "8.0.1"`), blocking consumers from upgrading to Babel 8.

## Fix

Widen the assertion to `^7.0.0-0 || ^8.0.0` — the range Babel uses for its own 7/8-compatible plugins ([babel/babel#17580](https://github.com/babel/babel/pull/17580)). No other change is needed: the transform uses only stable `@babel/types` builders and the standard visitor API.

## Verification

Under `@babel/core@8.0.1`, a `task(async (ms) => { await wait(ms); return ms + 1; })` transforms correctly to a `function*` generator task; the unpatched assert throws. Confirmed still working under 7.29.7.

### DIFF
--- a/packages/ember-concurrency/async-arrow-task-transform.js
+++ b/packages/ember-concurrency/async-arrow-task-transform.js
@@ -331,7 +331,7 @@ function cleanupUnusedTaskFactoryImports(programPath, usedTaskFactories) {
 }
 
 module.exports = declare((api) => {
-  api.assertVersion(7);
+  api.assertVersion('^7.0.0-0 || ^8.0.0');
 
   return {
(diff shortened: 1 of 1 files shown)


jiayingxu:
@machty it would be amazing to get this merged - it would unblock us from upgrading our application to Babel 8. Thanks for taking a look!

johanrd:
Closing, fixed by the duplicate https://github.com/machty/ember-concurrency/pull/613

